PR #57, opened 2025-02-12
[M-01] Approval Periods Are Implicitly Restricted

### LINKED ISSUE #35: [M-01] Approval Periods Are Implicitly Restricted

brozorec:
close #35 

The audit report suggests two ways of addressing this issue:
> Consider allowing accounts to specify approvals valid for longer than the maximum TTL value by limiting the TTL extension period to the maximum possible value. ...

and
> Alternatively, consider using the permanent storage for storing allowances.

We acknowledge the reasoning behind this suggestions, but decided to keep the current design of implicitly restricted approvals. The main reason is consistency with SAC (Stellar Asset Contract), where temporary storage is used for this data and one can't set a longer TTL than the max allowed. However, we introduce an explicit check/error for the case when `live_until_ledger > max_live_until_ledger` and explain this limitation in the docs.


contracts/token/fungible/src/test.rs
```diff
@@ -162,16 +162,30 @@ fn set_allowance_with_expired_ledger_fails() {
 
 #[test]
 #[should_panic(expected = "Error(Contract, #202)")]
+//#[should_panic]
```
ozgunozerk, line 165: should we remove this commented out part?

PR #84, opened 2025-02-25
Add explicit auth tests for fungible token functions

### LINKED ISSUE #69: 🎁 [Feature Request]: Assert `auths` in `fungible`
### What is the feature you would like to see?

All functions in `fungible` that require authorizations (`addr.require_auth()`) must have tests for the specific `auths` rather than just using `e.mock_all_auths()`. (Non-exhaustive) list of those functions:
- `approve`
- `transfer`
- `transfer_from`
- `burn`
- `burn_from`

As a reference, check [this example](https://developers.stellar.org/docs/build/smart-contracts/example-contracts/auth).

tinkerer-shubh:
Fixes #69 

This PR adds specific authorization tests for all functions in the fungible token contract that require authorization checks:
- approve
- transfer
- transfer_from
- burn
- burn_from

It also improves the documentation in the Mintable extension to clarify its intentional lack of authorization controls. (I may have oversimplified this part of the contribution, but I felt these changes were necessary to enhance understanding. If this change doesn’t align with the goals or isn’t needed, I’m happy to revert it! )

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/contracts/token/fungible/src/extensions/mintable/mod.rs
+++ b/contracts/token/fungible/src/extensions/mintable/mod.rs
@@ -46,8 +46,27 @@ pub trait FungibleMintable {
     /// [`crate::extensions::capped`], and check the `fungible-capped`
     /// example.
     ///
-    /// IMPORTANT: Please do not forget that, you probably will want to have
-    /// some authorization controls for minting tokens.
+    /// # Security Warning
+    ///
+    /// IMPORTANT: The base implementation of mint() intentionally lacks
+    /// authorization controls. You MUST implement proper authorization in
+    /// your contract. For example:
+    ///
+    /// ```rust
+    /// fn mint(&self, e: &Env, admin: Address, to: Address, amount: i128) {
+    ///     // 1. Require authorization from the admin
+    ///     admin.require_auth();
+    ///
+    ///     // 2. Verify admin has minting privileges (optional)
+    ///     let stored_admin = e.storage().instance().get(&ADMIN_KEY).unwrap();
+    ///     assert_eq!(admin, stored_admin, "unauthorized: not the admin");
+    ///
+    ///     // 3. Only then call the actual mint function
+    ///     crate::mintable::mint(e, &to, amount);
+    /// }
+    /// ```
+    ///
+    /// Failing to add proper authorization could allow anyone to mint tokens!
     fn mint(e: &Env, account: Address, amount: i128);
 }
 
--- a/contracts/token/fungible/src/extensions/mintable/mod.rs
+++ b/contracts/token/fungible/src/extensions/mintable/mod.rs
@@ -53,15 +53,12 @@ pub trait FungibleMintable {
     /// your contract. For example:
     ///
     /// ```rust
-    /// fn mint(&self, e: &Env, admin: Address, to: Address, amount: i128) {
-    ///     // 1. Require authorization from the admin
+    /// fn mint(&self, e: &Env, to: Address, amount: i128) {
+    ///     // 1. Verify admin has minting privileges (optional)
+    ///     let admin = e.storage().instance().get(&ADMIN_KEY).unwrap();
     ///     admin.require_auth();
     ///
-    ///     // 2. Verify admin has minting privileges (optional)
-    ///     let stored_admin = e.storage().instance().get(&ADMIN_KEY).unwrap();
-    ///     assert_eq!(admin, stored_admin, "unauthorized: not the admin");
-    ///
-    ///     // 3. Only then call the actual mint function
+    ///     // 2. Only then call the actual mint function
     ///     crate::mintable::mint(e, &to, amount);
     /// }
     /// ```
--- a/contracts/token/fungible/src/test.rs
+++ b/contracts/token/fungible/src/test.rs
@@ -434,6 +434,10 @@ fn update_with_insufficient_balance_panics() {
 }
 
 // Authorization Tests
+
+// Note: Invocation assertions are temporarily commented out while we investigate
+// an issue where auth entries are not being populated with function name and parameters
+// in the test environment.
 #[test]
 fn approve_requires_auth() {
     let e = Env::default();
--- a/contracts/token/fungible/src/test.rs
+++ b/contracts/token/fungible/src/test.rs
@@ -6,9 +6,9 @@ use soroban_sdk::{
     contract, symbol_short,
     testutils::{
         storage::{Instance, Persistent},
-        Address as _, Events, Ledger,
+        Address as _, AuthorizedFunction, Events, Ledger,
     },
-    vec, Address, Env, IntoVal,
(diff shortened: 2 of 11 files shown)


ozgunozerk (approved):
Thanks a lot for the tests! LGTM!

On `mint` security notes, we are already thinking about something to denote this security warning more explicitly (maybe embedding something into the function name, so the people who skip the inline comments also won't miss it). 

I think what you have proposed is better than what we currently have. So, thanks for that as well!

brozorec (changes requested):
Hey @tinkerer-shubh thanks for your contribution and sorry for the delayed review 🙏 
Catching `Error(Auth, InternalError)` by clearing the mocked auths is a viable strategy, but I'd recommend adopting a more rigorous one by asserting the specific authorizations [as shown](https://developers.stellar.org/docs/build/smart-contracts/example-contracts/auth#tests) in the docs:
```
    assert_eq!(
        env.auths(),
        [(
            // Address for which auth is performed
            user_1.clone(),
            // Identifier of the called contract
            contract_id.clone(),
            // Name of the called function
            symbol_short!("increment"),
            // Arguments used to call `increment` (converted to the env-managed vector via `into_val`)
            (user_1.clone(), 5_u32).into_val(&env)
        )]
    );
```

tinkerer-shubh:
@ozgunozerk @brozorec Thanks for the review!

I’ve made the requested changes—let me know what you think. Happy to make any further adjustments if needed!

contracts/token/fungible/src/extensions/mintable/mod.rs
```diff
@@ -41,8 +41,27 @@ pub trait FungibleMintable {
...
+    ///     // 1. Require authorization from the admin
+    ///     admin.require_auth();
+    ///
+    ///     // 2. Verify admin has minting privileges (optional)
+    ///     let stored_admin = e.storage().instance().get(&ADMIN_KEY).unwrap();
+    ///     assert_eq!(admin, stored_admin, "unauthorized: not the admin");
+    ///
+    ///     // 3. Only then call the actual mint function
+    ///     crate::mintable::mint(e, &to, amount);
+    /// }
```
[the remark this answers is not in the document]
tinkerer-shubh: thanks a lot for the suggestion! 

fixed this in a4eebb827b1884e4b11f99beffdb28be52f47f24

contracts/token/fungible/src/test.rs
```diff
@@ -431,3 +431,168 @@ fn update_with_insufficient_balance_panics() {
...
+                    &e,
+                    owner.clone().into_val(&e),
+                    spender.clone().into_val(&e),
+                    amount.into_val(&e),
+                    expiration_ledger.into_val(&e)
+                ]
+            ))
+        );
+    });
+}
```
brozorec, line 469: Hey, that's moving in the right direction, but structured this way makes the part after L448 non-relevant, because `approve()` will panic and execution stops there.

We still need to mock auths and the idea is to test if we get the right one for this particular invocation:
```
#[test]
fn approve_requires_auth() {
    let e = Env::default();
    e.mock_all_auths();
    let address = e.register(MockContract, ());
    let owner = Address::generate(&e);
    let spender = Address::generate(&e);
    let amount = 100;
    let expiration_ledger = 1000;

    e.as_contract(&address, || {
        approve(&e, &owner, &spender, amount, expiration_ledger);
    });

    let auths = e.auths();
    assert_eq!(auths.len(), 1);
    let (addr, invocation) = &auths[0];
    assert_eq!(addr, &owner);
    //assert_eq!(
        //invocation.function,
        //AuthorizedFunction::Contract((
            //address,
            //symbol_short!("approve"),
            //vec![
                //&e,
                //owner.clone().into_val(&e),
                //spender.clone().into_val(&e),
                //amount.into_val(&e),
                //expiration_ledger.into_val(&e)
            //]
        //))
    //);
}
```
`e.auths()` aren't accessible in the closure and must be asserted after. However, here smth unexpected happens, hence the commented chunk. I was expecting that the function name and params will be populated in the auth entry, which is not the case:
```
assertion `left == right` failed
  left: Contract((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Symbol(), Vec()))
 right: Contract((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAD2KM), Symbol(approve), Vec(Ok(Address(obj#11)), Ok(Address(obj#13)), Ok(I128(100)), Ok(U32(1000)))))
```

This behavior should be further examined and checked whether it's expected or it's a bug. Nevertheless, for the time this question gets cleared out, we still can assert `assert_eq!(addr, &owner);` which serves the test case.

brozorec (changes requested):

contracts/token/fungible/src/test.rs
tinkerer-shubh: Thanks a lot for the guidance! 

Addressed in https://github.com/OpenZeppelin/stellar-contracts/pull/84/commits/dba8ad4408028efa1f58a60fafdd30baaf01117f

tinkerer-shubh: > This behavior should be further examined to determine whether it's expected or a bug.


From my initial investigation, it seems like the function name and parameters should have been populated, so what we’re seeing may not be the expected behavior.

There are two possibilities: either something in the auth tests I wrote is causing this (though I wasn’t able to pinpoint the exact issue), or it could be related to how the test environment currently handles auth tests. Could this be an upstream issue?

brozorec: Indeed, I also suspect it's an upstream issue. Would you like to post an issue https://github.com/stellar/rs-soroban-sdk ?

contracts/token/fungible/src/test.rs
```diff
@@ -434,73 +435,71 @@ fn update_with_insufficient_balance_panics() {
...
-                ]
-            ))
-        );
     });
+
+    let auths = e.auths();
+    assert_eq!(auths.len(), 1);
+    let (addr, _invocation) = &auths[0];
+    assert_eq!(addr, &owner);
+    // assert_eq!(
```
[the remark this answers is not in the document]
brozorec: Yes, that's fine 👍 Maybe we can just have a short comment about the reason for leaving them commented. Other than that it's looking great, thank you 🙏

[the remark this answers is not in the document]
tinkerer-shubh: addressed in https://github.com/OpenZeppelin/stellar-contracts/pull/84/commits/06c771b336431e10df1c656bda7e37b4bc22cfb4

thanks for the review! 🚀

brozorec (approved):
Thanks for your contribution @tinkerer-shubh 🙏

PR #99, opened 2025-03-03
export constants and refactor the repo

### LINKED ISSUE #98: 🏗️ [Core Feature]: Export constants into its own crate

Constants like `INSTANCE_EXTEND_AMOUNT` are used in `fungible`, `non-fungible`, and in `pausable` (and will be used probably in more places in the future). 

These better be exported to its own crate for re-usability.

ozgunozerk:
Fixes #98

Draft because this is based on `non-fungible`. Will be ready to merge after `non-fungible` PR is merged.

I noticed some stuff about our repo while exporting the constants into its own crate:
- crate names are `openzeppelin_pausable`, `openzeppelin_fungible` etc. But the name `stellar` is not there. I think we have to change our crate names. I suggest we stick `stellar` in between like: `openzeppelin_stellar_pausable`, etc. There may be and will be other openzeppelin crates out there, and we should somehow use the name `stellar` for this. 
- inside `contracts/token` folder, we had only `fungible`, but soon we will have `non-fungible` and others. I suggest we rename this into `tokens` to be consistent with `contracts` and `utils` as well.


brozorec:
I'm also in favor for including `stellar` in the name, but I'd suggest we switch from exporting separately every member from the workspace and have one `openzeppelin_stellar` which gives access to individual modules: `openzeppelin_stellar::pausable`, `openzeppelin_stellar::fungible`, etc. That looks visually more natural, leverages Rust mod approach and follows the common practice.

ozgunozerk:
We had these following arguments (from an internal discussion):

Monorepo approach is quite common nowadays, especially amongst Rust projects.
But the reasons monorepo approach (no root level `lib.rs`) is followed due to the following (AFAIK):
- If the entire monorepo is exposed as a single crate, any change in one module could trigger recompilation of the entire codebase.
- Having separate crates allows independent versioning

I agree that this approach looks better though:
> openzeppelin_stellar::pausable, openzeppelin_stellar::fungible, etc. That looks visually more natural

ozgunozerk:
closing in favor of #107 (creating a new PR was easier than resolving conflicts)

PR #103, opened 2025-03-07
Non-fungible tests

brozorec:
Added some test cases while reviewing, so pushing them here to be used if considered useful.


ozgunozerk (approved):
Thanks!

PR #108, opened 2025-03-13
Enumerable non fungible

### LINKED ISSUE #80: 🏗️ [Core Feature]: Implement `Enumerable` for `Non-Fungible`

This will allow contracts to publish their full list of NFTs and make them discoverable.

abrarabedin001:
Fixes #80 

Key features:

- Maintains separate lists for each owner's tokens
- Keeps track of all tokens in the contract
- Properly updates total supply during minting and burning
- All storage functions are in the extension file
- No changes to the base NFT implementation

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/non-fungible/src/extensions/enumerable.rs
+++ b/packages/tokens/non-fungible/src/extensions/enumerable.rs
@@ -1,0 +1,82 @@
+use soroban_sdk::{contracttype, Address, Env, Vec};
+use crate::storage::{StorageKey, owner_of};
+
+/// Extension trait that provides enumeration capabilities for NFTs
+pub trait NonFungibleEnumerable {
+    /// Returns the total number of tokens stored by the contract.
+    fn total_supply(e: &Env) -> u128;
+
+    /// Returns a list of token IDs owned by `owner`.
+    fn tokens_of_owner(e: &Env, owner: Address) -> Vec<u128>;
+
+    /// Returns a list of all token IDs in the contract.
+    fn all_tokens(e: &Env) -> Vec<u128>;
+}
+
+/// Storage key for enumerable extension data
+#[contracttype]
+pub enum EnumerableStorageKey {
+    /// Key for storing token IDs owned by an address
+    TokensOfOwner(Address),
+    /// Key for storing all token IDs
+    AllTokens,
+}
+
+/// Returns the total number of tokens stored by the contract.
+pub fn total_supply(e: &Env) -> u128 {
+    e.storage().instance().get(&StorageKey::TotalSupply).unwrap_or(0)
+}
+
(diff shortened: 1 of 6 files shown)


abrarabedin001:
@ozgunozerk, can you kindly review this? This time, I tried to fix the previous mistakes.

ozgunozerk:
Thanks a lot!

The issues I've mentioned earlier are indeed addressed. 

Moving on to the design decisions and business logic:
- I think if we provide override methods for `burn`, `mint`, and `transfer` -> it would improve the DevX a lot. To be more concrete, I'm suggesting the `Enumerable` trait should have its own `burn`, `mint`, and `transfer` methods, and we should explain in the inline comments that, `Enumerable` extension is overriding these behaviors, due to newly added `total_supply` functionality. So, the developer shouldn't use `non_fungible::mint` but use `non_fungible::enumerable::mint` if they want to use `Enumerable`. This is also how it's done in Stylus Contracts as well, see: https://docs.openzeppelin.com/contracts-stylus/0.1.0/erc721-enumerable
- Although hierarchically it looks good, I'd advice against and additional `helpers` mod. Because, then it would be: `stellar_non_fungible::enumerable::helpers::burn` -> it's just too long... Having all the methods under the `Enumerable` trait is fine.
- This implementation stores the "IDs of the tokens owned by an `Owner`" in a Vector. And utilizes binary search for removing the token. This is costly. There could be, and I believe there already is a better implementation that offers better time complexity. A good read may be: https://www.rareskills.io/post/erc-721-enumerable.
- Possible overflows and underflows are not checked.

These are the major points I saw from the first glance. After these are resolved, we should be mostly good to go.

I really appreciate the effort @abrarabedin001, this task you took is not a simple one!

I hope your enthusiasm is intact, and would love to see this develop more <3

ozgunozerk (changes requested):
@abrarabedin001 , we also started to think about `Consecutive`, and similar problems are present there as well (overriding methods, and not being compatible with `burnable` or `mintable`, and maybe even the `update` function has to be overridden). 

We will most probably finalize our design decisions and strategies on Wednesday (maybe trait, or maybe generics based solution, or maybe something else). 

Till that time, I suggest that we put a hold on this issue, because it could be that, what I suggested may get obsolete on Wednesday as well.

abrarabedin001:
Hi Guys, it was a great learning experience. I will keep an eye out for your repo. I have a lot going on, so I will close it for now. I will contribute once I am more free.

ozgunozerk:
@abrarabedin001, thanks a lot, we also enjoyed the discussion 👍 
Feel free to take a look at our `good-first-issues` if you ever want to contribute in the future.
Hope to see you again, and best of luck!

PR #117, opened 2025-03-18
Upgrades and migrations

### LINKED ISSUE #83: 🏗️ [Core Feature]: Upgradability

@brozorec could you fill in the details for this one? Feel free to pick `small`, `medium`, or `big` from the labels as you see fit!

brozorec:
Fixes #83 

This PR proposes an upgradeability framework with optional migration and rollback utilities. It makes usage of derive macros to streamline the implementation process. Follows the specs of [Upgradeable Contracts SEP](https://github.com/stellar/stellar-protocol/discussions/1670).

#### PR Checklist

- [x] Tests
- [ ] Documentation


Cargo.toml
```diff
@@ -1,7 +1,10 @@
 [workspace]
 resolver = "2"
 members = [
-    "examples/*",
+    "examples/fungible-capped",
+    "examples/fungible-pausable",
+    "examples/fungible-token-interface",
+    "examples/upgradeable/*",
```
ozgunozerk, line 7: why not just `examples/*`?

examples/upgradeable/v1/src/contract.rs
```diff
@@ -0,0 +1,35 @@
+use soroban_sdk::{
```
ozgunozerk, line 1: since we have many examples for the upgradeable, a comment on top of this file, explaining its purpose and the diff between `v2` (or how they work together) would be good

ozgunozerk: the same goes for `v2` file in examples

examples/upgradeable/v1/src/test.rs
```diff
@@ -0,0 +1,46 @@
...
+
+fn install_new_wasm(e: &Env) -> BytesN<32> {
+    e.deployer().upload_contract_wasm(contract_v2::WASM)
+}
+
+type MigrationData = Data;
+type RollbackData = ();
+
+#[test]
+fn test_upgrade() {
```
ozgunozerk, line 24: I'll give another look to this test after reading the sourcecode (started reviewing first from test files). And currently, I don't have a good idea on what's happening in here. There may be other people who will take a look first at test files to quickly grasp what this code is about. Maybe we can put some inline comments here and there in this test, explaining what's happening and what's the expected behavior for the asserts.

packages/contract-utils/upgradeable-macros/src/derive.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Procedural macro implementation for `#[derive(Upgradeable)]`.
+///
+/// This function generates the implementation of the `Upgradeable` trait for a
+/// given contract type, enabling the contract to be upgraded by replacing its
+/// WASM bytecode.
+///
+/// If the contract struct is annotated with the `#[migratable]` attribute, this
+/// macro will also generate an implementation of the `Migratable` trait, wiring
+/// up the migration and rollback logic based on the `MigratableInternal` trait
+/// provided by the user.
```
ozgunozerk, line 14: what happens if `#[migratable]` is there, but user does not provide `MigratableInternal`?

packages/contract-utils/upgradeable-macros/src/derive.rs
```diff
@@ -0,0 +1,98 @@
...
+                e.deployer().update_current_contract_wasm(new_wasm_hash);
+            }
+        }
+
+        #migratable
+    }
+}
+
+// Internal helper to derive the `Migratable` trait
+fn derive_migratable(name: &Ident) -> proc_macro2::TokenStream {
```
ozgunozerk, line 71: from what I know, macros have access to the other macros in their list. So, this should be also possible:
`#[derive(Upgradable, Migratable)]`
and how it may work like this:
- if there is only `Upgradable`, it will work as normal
- if there is both `Upgradable` and `Migratable` -> Upgradable will consume `Migratable`, and change its behavior accordingly (will also implement Migratable)
- if there is only `Migratable` -> it should give an error

packages/contract-utils/upgradeable-macros/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+/// 1. Derives Upgradeable a) implements the interface; requires only the auth
+///    to be defined b) sets wasm version by taking the version from Cargo.toml
+///
+/// 2. Optionally derives Migratable when migration and rollback are defined.
+///
+///
+/// Example:
+/// ```rust,ignore
+/// #[derive(Upgradeable)]
+/// #[migratable]
```
ozgunozerk, line 10: I couldn't see the code for the scenario where the user only provides `#[migratable]`.

I think there should be a `migratable` macro as well, and give an error, saying that `migratable` shouldn't be used without `upgradable`. 

And, `upgradable` can consume the macro `#[migratable]` (effectively, delete it), so that, `migratable` macro inner function won't be called when `#[derive(Upgradable)]` is present. 

I have another personal following this approach, can give examples if wanted.

packages/contract-utils/upgradeable-macros/src/derive.rs
ozgunozerk: This suggestion above is subjective. I think it will be better to have them together as a single-liner, but let me know if you prefer the current approach 👍

packages/contract-utils/upgradeable/src/storage.rs
```diff
@@ -0,0 +1,106 @@
+use soroban_sdk::{contracttype, panic_with_error, symbol_short, Env, Symbol};
+
+use crate::upgradeable::UpgradeableError;
+
+pub const UPGRADE_KEY: Symbol = symbol_short!("UPGRADE");
+
+/// Represents the current upgrade state of the contract. Used to determine if
+/// migration or rollback operations are allowed.
+#[contracttype]
+pub enum UpgradeState {
```
ozgunozerk, line 10: I really like this approach of keeping an enum to track states. If, in the future, we want to further optimize this cost, there is a way to enforce the state using types in compile time, have a look at: https://docs.rs/state-shift/latest/state_shift/

It also hides the underlying generics from the end-developer. However, since stellar/soroban does not support generics at some places, I'm not sure whether this will work or not.

May worth trying though at a later stage, not a blocker for this PR

packages/contract-utils/upgradeable/src/upgradeable.rs
```diff
@@ -0,0 +1,112 @@
...
+///    - Versioning management
+///
+/// 2. Framework-assisted usage – When using the lightweight upgrade framework
+///    provided in this module, you should NOT manually implement this trait.
+///    Instead:
+///    - Derive it using `#[derive(Upgradeable)]`
+///    - Provide access control by implementing [`UpgradeableInternal`] with
+///      your custom logic
+#[contractclient(name = "UpgradeableClient")]
+pub trait Upgradeable {
```
ozgunozerk, line 20: should this have a bound `Upgradable: UpgradableInternal`? 
I believe, you didn't put it because someone may implement it in a custom way that wouldn't require `UpgradableInternal`. 

But couldn't we use the same argument for `Migratable: MigratableInternal`?

ozgunozerk (commented):
Great job!

examples/upgradeable/upgrader/src/contract.rs
```diff
@@ -0,0 +1,54 @@
...
+        operator: Address,
+        old_wasm_hash: BytesN<32>,
+        rollback_data: soroban_sdk::Vec<Val>,
+    ) {
+        let contract_client = UpgradeableClient::new(&env, &contract_address);
+
+        // The types of the arguments to the migrate function are unknown to this
+        // contract, so we need to call it with invoke_contract.
+        env.invoke_contract::<()>(&contract_address, &ROLLBACK, rollback_data);
+        contract_client.upgrade(&old_wasm_hash, &operator);
```
ozgunozerk, line 52: `contract_client.upgrade` is used in downgrade function. It kind of bothered me to see `upgrade()` is used within `downgrade()` implementation

packages/contract-utils/upgradeable/src/upgradeable.rs
```diff
@@ -0,0 +1,112 @@
+use soroban_sdk::{contractclient, contracterror, Address, BytesN, Env, FromVal, Val};
```
ozgunozerk, line 1: should we emit an event for `upgrade` and `downgrade`? 
It may be useful to have this info available on block explorers, wdyt?

ozgunozerk, line 1: should we emit an event for `upgrade` and `downgrade`? 
It may be useful to have this info available on block explorers, wdyt?

packages/contract-utils/upgradeable-macros/src/lib.rs
brozorec: `#[migratable]` is an attribute of the derive macro and can't be used standalone https://github.com/OpenZeppelin/stellar-contracts/blob/e44e41dc578024454d715e27ac7732df7e8ad610/packages/contract-utils/upgradeable-macros/src/lib.rs#L40

Nevertherless, decided to follow your idea of having  `#[derive(Migratable)]`.

packages/contract-utils/upgradeable/src/upgradeable.rs
brozorec: There's a [system event emitted](https://github.com/stellar/rs-soroban-env/blob/e3feadbc570b9fb88769dbea26e83a52f48cca2e/soroban-env-host/src/events/system_events.rs#L16) when updating a contract executable code, making additional event unnecessary here.

examples/upgradeable/upgrader/src/contract.rs
brozorec: Yeah, I see. I also thought about it, but decided that there's no need to define another function that exactly the same as the first one, just with a different name. Maybe, it's better to rename this one to `rollback_and_upgrade`.

packages/contract-utils/upgradeable/src/upgradeable.rs
brozorec: Yes, that's correct about `UpgradableInternal`, but `Migratable` and `MigratableInternal` are interconnected as they share the same accoc types. If someone wants something that doesn't fit within the provided framework, they simply don't need those traits.

packages/contract-utils/upgradeable/src/upgradeable.rs
ozgunozerk: good to know, thanks!

ozgunozerk (approved):

PR #121, opened 2025-03-19
refactor: add find_even_by_symbol function

### LINKED ISSUE #111: Refactor `event_assertion` crate
`packages/test-utils/event-assertion/src/lib.rs` file has many duplications for the functions it expose. 

We should refactor it, and re-use some of the common parts, and reduce the duplication across the functions.

One additional good idea may be to create `mod`s, namely: 
- fungible
- non-fungible

and organize the functions accordingly (for example, currently we have `assert_fungible_transfer` and `assert_non_fungible_transfer`, which doesn't look good).

YanVictorSN:
Fixes #111  

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ozgunozerk:
That's good! I like the changes 👍 

However, I think we can do further categorization. 

What I have in mind, as I mentioned in the issue #111 , 

there are `fungible` and `non-fungible` variations of the same events. We should reduce the duplication across these if possible. 

If not possible, then, at worst, we can split them into 2 mods, like:

```rust

/* common functions, constants, whatever, are here */

mod fungible {
   /* fungible related functions should be there
}

mod non_fungible {
   /* non fungible related functions should be there
}
```

YanVictorSN:
@ozgunozerk Thank you! Is it possible to merge this one? I’ll open another one with more refactoring. I'm a Rust novice, so it may take a while.

ozgunozerk:
What I'm suggesting is easier than what you accomplished, don't underestimate yourself <3 

Soon `nft-mintable` PR will also be merged, which tweaks this file a bit. It makes more sense to merge this one after `nft-mintable` PR.

And also, I prefer to have this `event_assertion` module refactored in a single go, since the scope is quite small. 

If there are any unclear points, I can submit a draft-pr, showcasing what I have in mind. It's essentially just about moving stuff around, no new code is required :)

YanVictorSN:
Thank you for the words! <3

Checking the code, the only difference I see between the paired functions is that one has `amount: i128` and the other has `token_id: u32`.

I'm thinking of creating an `enum` for the parameter and a function that can handle both `fungible` and `non-fungible` events. Does this align with what you're thinking?

ozgunozerk:
If you can find a way to reduce the duplication across `fungible` and `non_fungible`, that would be great. But it's not required.

What I have in mind is this:

```rust
use soroban_sdk::{symbol_short, testutils::Events, Address, Env, IntoVal, Symbol, Val, Vec};

pub struct EventAssertion<'a> {
    env: &'a Env,
    contract: Address,
}

impl<'a> EventAssertion<'a> {
    pub fn new(env: &'a Env, contract: Address) -> Self {
        Self { env, contract }
    }
}

mod fungible {
    use super::*;

    impl<'a> EventAssertion<'a> {
        // this was `assert_fungible_transfer`
        pub fn assert_transfer(&self, from: &Address, to: &Address, amount: i128) { 
           /* body of the function */
        }
        
        /* other fungible related functions here */        
}

mod non_fungible {
    use super::*;
    
    impl<'a> EventAssertion<'a> {
        // this was `assert_non_fungible_transfer`
        pub fn assert_transfer(&self, from: &Address, to: &Address, amount: i128) {
           /* body of the function */
        }
    }
}
```

and these functions will be used elsewhere like this:
```rust
use event_assertion::fungible::{transfer, mint, burn};
```

instead of 
```rust
use event_assertion::{fungible_burn, fungible_transfer, fungible_mint};
```

Does it makes sense now?

YanVictorSN:
Yes. Now make sense. Thank you!

YanVictorSN:
I tried the way you suggested, but Rust threw an error about the duplication of definitions even in different modules. 

Is there a way to work around that? 

Should I just separate them into different modules without renaming?

So far has 259 lines

ozgunozerk:
Ah yes, you're right. 

Since they are both implementing `EventAssertion` directly, the same named methods will cause a problem.

To workaround that, we have to provide different traits, like `FungibleEventAssertion` and `NonFungibleEventAsserion`.

And I'm not sure how much value that will bring us, the complexity may not worth it for this simple module. 

Let's keep as is without the separation 👍

YanVictorSN:
That's okay. 

I learned a lot! Thank you! 

So I will wait for the `nft-mintable` to get merged, then I’ll resolve the conflicts. 

In the meantime, I'll look for other good first issues.

ozgunozerk (approved):

PR #124, opened 2025-03-23
NFT consecutive

### LINKED ISSUE #82: 🏗️ [Core Feature]: Implement `Consecutive` for `Non-Fungible`

Allowing contracts to `mint`, `transfer`, or `burn` batches of tokens for scalability.

brozorec:
Fixes #82 

#### PR Checklist

- [x] Tests
- [ ] Documentation


packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -0,0 +1,341 @@
...
+pub fn owner_of(e: &Env, token_id: TokenId) -> Address {
+    let max = sequential::next_token_id(e);
+    let is_burned =
+        e.storage().persistent().get(&StorageKey::BurnedToken(token_id)).unwrap_or(false);
+
+    if token_id >= max || is_burned {
+        panic_with_error!(&e, NonFungibleTokenError::NonExistentToken);
+    }
+
+    //e.storage().persistent().extend_ttl(&key, OWNER_TTL_THRESHOLD,
```
ozgunozerk, line 42: commented out line

packages/tokens/non-fungible/src/extensions/consecutive/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+use crate::{NonFungibleToken, TokenId};
+
+/// Consecutive Marker Trait for Non-Fungible Token
+///
+/// # Notes
+///
+/// The `consecutive` extension provides its own business logic for creating and
+/// destroying tokens. Therefore, this trait is INCOMPATIBLE with the
+/// `Mintable`, `Burnable`, and `Enumerable` extensions.
+pub trait NonFungibleConsecutive: NonFungibleToken<ContractType = Consecutive> {}
```
ozgunozerk, line 50: I think it will be a good idea to include `batch_mint` here as a method. My reasoning is this:
when someone implements this trait on their contract, seeing that they have to implement `batch_minting` makes sense. Otherwise, why are they even implementing this?

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -0,0 +1,341 @@
...
+///
+/// Failing to add proper authorization could allow anyone to mint tokens!
+pub fn batch_mint(e: &Env, to: &Address, amount: TokenId) -> TokenId {
+    let next_id = sequential::increment_token_id(e, amount);
+
+    e.storage().persistent().set(&StorageKey::Owner(next_id), &to);
+
+    increase_balance(e, to, amount);
+
+    let last_id = next_id + amount - 1;
```
ozgunozerk, line 98: I'm confused on why do we add `amount` once more? 
`next_id` is already = `current_id` + `amount`
so, in this code, `last_id` = `current_id` + `amount` + `amount` - 1

packages/tokens/non-fungible/src/storage.rs
```diff
@@ -373,3 +405,36 @@ pub fn check_spender_approval(e: &Env, spender: &Address, owner: &Address, token
...
+/// Low-level function for decreasing the balance of `to`, without handling
+/// authorization. Balance cannot go negative.
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `to` - The address whose balance gets decreased.
+/// * `amount` - The amount by which the balance gets decreased.
+pub fn decrease_balance(e: &Env, from: &Address, amount: TokenId) {
+    let new_balance = balance(e, from).saturating_sub(amount);
```
ozgunozerk, line 438: In case of this function to be called on its own, I suggest we replace `saturating_sub` with `checked_sub` and generate an error where the case it would fall below `0`

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -0,0 +1,341 @@
...
+    } else {
+        // Burning: `to` is None
+        e.storage().persistent().remove(&StorageKey::Owner(token_id));
+
+        e.storage().persistent().set(&StorageKey::BurnedToken(token_id), &true);
+    }
+}
+
+/// Low-level function that sets owner of `token_id` to `to` if the token
+/// exists, without handling authorization.
```
ozgunozerk, line 325: maybe also mention `if the token exists, and it doesn't have an owner at the moment` or smt along those lines

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -0,0 +1,341 @@
...
+/// * `e` - The environment reference.
+/// * `to` - The owner's address.
+/// * `token_id` - The identifier of the token being set.
+pub fn set_owner_for(e: &Env, to: &Address, token_id: TokenId) {
+    let max = sequential::next_token_id(e);
+    let has_owner = e.storage().persistent().has(&StorageKey::Owner(token_id));
+    let is_burned =
+        e.storage().persistent().get(&StorageKey::BurnedToken(token_id)).unwrap_or(false);
+
+    if token_id < max && !has_owner && !is_burned {
```
ozgunozerk, line 338: I'm a bit puzzled. `set_owner_for` only sets the owner, if there is no active owner for that token in the storage, so far so good. 
But, the `update` function, does not remove the current owner? I know that the tokens in between the range does not have an owner. But what if the tokens at the beginning or at the end are to be transferred? Will this still work?

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
brozorec: it will be better handled with all the others in https://github.com/OpenZeppelin/stellar-contracts/issues/129

packages/tokens/non-fungible/src/extensions/consecutive/mod.rs
brozorec: They are not obliged to implement `batch_mint` as a public function, batch-minting will be done in many cases from the constructor. Here comes the same reasoning as for the utility of both mintable traits I expressed in a dm.

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
brozorec: > `next_id` is already = `current_id` + `amount`

No `next_id` is the first "free" ID, it's basically `first_id` from the batch. Good point, will rename it to avoid confusion.

ozgunozerk:
Do we have a similar limitation? May be worth to note it either case. 

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Consecutive.sol#L19-L20

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
brozorec: > But what if the tokens at the beginning or at the end are to be transferred? Will this still work?

Let's see how that goes for
[token{0}, token{1}, ..., token{n}]

1. initial batch mint: only set token{0} = ownerA
2. on transfer token{0} where `Some(from), Some(to)`: set token{0} = onwerB and set next token{1} = ownerA
3. on transfer token{n} where `Some(from), Some(to)`: only set token{n} = ownerB, because `set_owner_for` does nothing in that case.

brozorec:
> Do we have a similar limitation? May be worth to note it in either case.
> 
> https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/extensions/ERC721Consecutive.sol#L19-L20

Only `batch_mint` must be used in either case and we need to mention it 👍

ozgunozerk (commented):
💯

ozgunozerk (approved):

Issue #123, opened 2025-03-20
🐞 [Bug]: Event Assertion create only handles the first error of a kind


ozgunozerk:
### What happened?

Take a look at these lines: https://github.com/OpenZeppelin/stellar-contracts/blob/main/packages/tokens/non-fungible/src/test.rs#L444

I've tested the `minting` and `counter` part, and `tokens_id`s are indeed `0` and `1`. However, although there are 2 mint events, the `event_assert` crate only deals with the first `mint` event every time (due to `find()` logic used inside of `event_assert` crate). 

### Expected behavior

We should somehow fix this, and allow `event_assert` crate to parse the next event of the same kind for each call. 

The most basic approach would be, deleting the parsed event from the queue. If the actual `events` queue is not be touched, we should:
1. clone the events for `event_assert` crate
2. delete the parsed event from the cloned list

elisafalk:
Hello @ozgunozerk . I'm interested in working on this issue.

ozgunozerk:
You got it! Have fun and don't hesitate to ask questions if you have any

elisafalk:
Hey @ozgunozerk I have send the PR here. https://github.com/OpenZeppelin/stellar-contracts/pull/145 . Can you please do a review?

YanVictorSN:
Hi @ozgunozerk! I'm trying to resolve this issue, but I’m having some problems.

The bug we want to solve is that `find()` only gets the first event and we don’t have a way to track if the event has been read or not, so it always reads the same one.

Following your suggestion, I tried adding a way to track the event by the index. But I’m not sure if it’s the best approach because I have to make the `EventAssertion` mutable, and I’m not sure if that’s a safe approach.

So, I have a few questions:

When I want to keep track of the events, should I add a way to track that in the `EventAssertion` struct, or should I try to track it inside the implementation?

Is it ok to make `EventAssertion` `mutable`, or do we want another way to keep track of that?

ozgunozerk:
Hi @YanVictorSN !

> Is it ok to make EventAssertion mutable

EventAssertion is used only internally as a test utility, I don't see any reason to not make it mutable. In fact, if we are to keep a list/index to track events, and modify it, making it mutable is more logical.

> When I want to keep track of the events, should I add a way to track that in the EventAssertion struct, or should I try to track it inside the implementation?

[...]

PR #145, opened 2025-04-06
 Fix: Enable EventAssertion to process multiple events of the same type

elisafalk:
# Fix: Enable EventAssertion to process multiple events of the same type

- Introduced RefCell for processed_events to track which events have been processed
- Modified find_event_by_symbol to return the next unprocessed event of the same type
- Added event_cache to optimize event retrieval and reduce redundant calls
- Ensured all Soroban Vec operations use proper SDK methods (push_back, Vec::new with env)
- Implemented refresh_events method to reset processed events tracking
- Maintained existing API signatures to ensure test compatibility
- Fixed issue where event assertions could only handle the first event of a kind

This PR fixes the issue where the event_assert crate only deals with the first event of a kind due to the find() logic. The solution tracks processed events to ensure subsequent calls to assertion methods can find and validate the next event of the same type.

Fixes https://github.com/OpenZeppelin/stellar-contracts/issues/123 

#### PR Checklist

- [x] Tests - All existing tests pass with the fix applied
- [ ] Documentation - No documentation changes needed as the API remains unchanged

### DIFF
--- a/packages/test-utils/event-assertion/src/lib.rs
+++ b/packages/test-utils/event-assertion/src/lib.rs
@@ -2,20 +2,35 @@
 
 use soroban_sdk::{symbol_short, testutils::Events, Address, Env, IntoVal, Symbol, Val, Vec};
 use stellar_non_fungible::TokenId;
+use std::cell::RefCell;
 
 pub struct EventAssertion<'a> {
     env: &'a Env,
     contract: Address,
+    processed_events: RefCell<Vec<(u32, u32)>>,
+    event_cache: RefCell<Option<Vec<(Address, Vec<Val>, Val)>>>,
 }
 
 impl<'a> EventAssertion<'a> {
     pub fn new(env: &'a Env, contract: Address) -> Self {
-        Self { env, contract }
+        Self { 
+            env, 
+            contract,
+            processed_events: RefCell::new(Vec::new(env)),
+            event_cache: RefCell::new(None),
+        }
+    }
+
+    fn get_all_events(&self) -> Vec<(Address, Vec<Val>, Val)> {
+        let mut cache = self.event_cache.borrow_mut();
+        if cache.is_none() {
+            *cache = Some(self.env.events().all());
+        }
+        cache.clone().unwrap()
     }
 
     fn find_event_by_symbol(&self, symbol_name: &str) -> Option<(Address, Vec<Val>, Val)> {
-        let events = self.env.events().all();
-
+        let events = self.get_all_events();
         let target_symbol = match symbol_name {
             "transfer" => symbol_short!("transfer"),
             "mint" => symbol_short!("mint"),
@@ -24,11 +39,35 @@ impl<'a> EventAssertion<'a> {
             _ => Symbol::new(self.env, symbol_name),
         };
 
-        events.iter().find(|e| {
-            let topics: Vec<Val> = e.1.clone();
+        for (event_index, event) in events.iter().enumerate() {
+            let topics: Vec<Val> = event.1.clone();
+            if topics.is_empty() {
+                continue;
+            }
+            
             let topic_symbol: Symbol = topics.first().unwrap().into_val(self.env);
-            topic_symbol == target_symbol
-        })
+            if topic_symbol == target_symbol {
+                let event_key = (event_index as u32, 0);
+                let mut processed_events = self.processed_events.borrow_mut();
+                if !processed_events.contains(&event_key) {
+                    processed_events.push_back(event_key);
+                    return Some(event.clone());
+                }
+            }
+        }
+        
+        None
+    }
+
+    pub fn assert_event_count(&self, expected: usize) {
+        let events = self.env.events().all();
+        assert_eq!(
+            events.len() as usize,
+            expected,
+            "Expected {} events, found {}",
+            expected,
+            events.len()
+        );
     }
 
     pub fn assert_fungible_transfer(&self, from: &Address, to: &Address, amount: i128) {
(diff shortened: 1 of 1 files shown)


packages/test-utils/event-assertion/src/lib.rs
```diff
@@ -2,20 +2,35 @@
 
 use soroban_sdk::{symbol_short, testutils::Events, Address, Env, IntoVal, Symbol, Val, Vec};
 use stellar_non_fungible::TokenId;
+use std::cell::RefCell;
 
 pub struct EventAssertion<'a> {
     env: &'a Env,
     contract: Address,
+    processed_events: RefCell<Vec<(u32, u32)>>,
+    event_cache: RefCell<Option<Vec<(Address, Vec<Val>, Val)>>>,
```
ozgunozerk, line 11: Why there is `Option` in `event_cache`? Can’t we handle the `None` case with checking for if the `Vec` is empty?

packages/test-utils/event-assertion/src/lib.rs
```diff
@@ -24,11 +39,35 @@ impl<'a> EventAssertion<'a> {
...
+            let topics: Vec<Val> = event.1.clone();
+            if topics.is_empty() {
+                continue;
+            }
+            
             let topic_symbol: Symbol = topics.first().unwrap().into_val(self.env);
-            topic_symbol == target_symbol
-        })
+            if topic_symbol == target_symbol {
+                let event_key = (event_index as u32, 0);
```
ozgunozerk, line 50: what does `0` corresponds to here exactly? Why are we assigning `0` to every event?

packages/test-utils/event-assertion/src/lib.rs
```diff
@@ -285,4 +308,21 @@ impl<'a> EventAssertion<'a> {
...
+        for expected_id in expected_ids {
+            self.assert_non_fungible_mint(to, *expected_id);
+        }
+    }
+
+    pub fn assert_non_fungible_transfers_ordered(&self, from: &Address, to: &Address, expected_ids: &[TokenId]) {
+        for expected_id in expected_ids {
+            self.assert_non_fungible_transfer(from, to, *expected_id);
+        }
+    }
```
ozgunozerk, line 327: None of these are used anywhere?

ozgunozerk (changes requested):
@elisafalk thanks for the contribution, there are few things that didn't make much sense to me, and tbh I'm suspecting this PR was created by an LLM, which is completely fine by the way (we are also using LLMs), but one has to supervise these LLM outputs. 

Have you tried to uncomment the lines for the tests (mentioned in the issue) to see whether this changes actually work? Because I haven't see any test files under modified files for this PR

packages/test-utils/event-assertion/src/lib.rs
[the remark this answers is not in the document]
elisafalk: The RefCell was initially included to enable interior mutability patterns that would allow the struct methods to use &self instead of &mut self, making API more well designed. But you are right as it makes the code harder to understand for newcomers. If we remove RefCell and switch to using &mut self for mutable operations, this will require changes to all test files where EventAssertion is used, since they'll need to declare the instance as mut.

elisafalk: The Option wrapper was use to initialize the cache only when needed. Although I can switch to empty Vec checks if preferred for consistency.

packages/test-utils/event-assertion/src/lib.rs
elisafalk: The refresh_events() and ordered assertion methods were included to provide a way to reset state between two scenarios,support commom test patterns for verifying event sequences and maintain consistency with similar test utiities.

elisafalk:
Hello @ozgunozerk . Thanks a lot for your review and feedback. Initially I had 2 approaches to solve this issue. 1) &mut self and make changes to the test files where EventAssertion is used and 2) Use RefCell to mutate the contents inside an immutable reference. I took the second approach.  
Regarding LLM- Yes, it's partially vibe coding since I'm quite new to open source and also to Rust. I just leveraged AI to get some basic boilerplate done, but I'm learning a lot along the way. I hope that won't be an issue moving forward and don't worry I have full control over the issue. I would appreciate your feedback and guidance. If you want I can submit an updated PR over the changes you want me to make.

packages/test-utils/event-assertion/src/lib.rs
ozgunozerk: Yes, for now I don't see a benefit of `Option` here. We can proceed with empty Vec

packages/test-utils/event-assertion/src/lib.rs
ozgunozerk: Adding it later wouldn't be a big problem, so let's keep it as simple as we can for now.

packages/test-utils/event-assertion/src/lib.rs
ozgunozerk: You could have at least read the question and the answer it provided before posting it here :) 
I'm not here to have a convo with an LLM, but to provide help to people who want to contribute.

packages/test-utils/event-assertion/src/lib.rs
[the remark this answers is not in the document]
ozgunozerk: `RefCell` makes sense for keeping the API as is. But I think we don't need to hide the `mut`, in fact, not hiding it also signals what's happening under the hood. Additionally, this is just testing, so it's not a breaking change from the end-user's perspective. 
I'd vote for getting rid of the `RefCell`.

PR #176, opened 2025-04-14
fix: make event_assert crate handle multiple events in sequence

### LINKED ISSUE #123: 🐞 [Bug]: Event Assertion create only handles the first error of a kind

YanVictorSN:
Fixes #123 

This PR changes how the `find_event_by_symbol` from `EventAssertion` handles multiple events in sequence. To achieve this, we introduced a `HashSet` called `processed_events` to keep track of the indices of the events that have already been read.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/tokens/non-fungible/src/test.rs
+++ b/packages/tokens/non-fungible/src/test.rs
@@ -55,7 +55,7 @@ fn approve_for_all_works() {
         let is_approved = Base::is_approved_for_all(&e, &owner, &operator);
         assert!(is_approved);
 
-        let event_assert = EventAssertion::new(&e, address.clone());
+        let mut event_assert = EventAssertion::new(&e, address.clone());
         event_assert.assert_event_count(1);
         event_assert.assert_approve_for_all(&owner, &operator, 1000);
     });
@@ -85,7 +85,7 @@ fn revoke_approve_for_all_works() {
         let is_approved = Base::is_approved_for_all(&e, &owner, &operator);
         assert!(!is_approved);
 
-        let event_assert = EventAssertion::new(&e, address.clone());
+        let mut event_assert = EventAssertion::new(&e, address.clone());
         event_assert.assert_event_count(1);
         event_assert.assert_approve_for_all(&owner, &operator, 0);
     });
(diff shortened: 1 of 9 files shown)


ozgunozerk:
Very straight forward, and works! Love it <3

packages/tokens/non-fungible/src/test.rs
```diff
@@ -437,7 +437,7 @@ fn test_counter_works() {
         let token_id1 = Base::sequential_mint(&e, &owner);
         let _token_id2 = Base::sequential_mint(&e, &owner);
```
ozgunozerk, line 438: @YanVictorSN can you also fix this and uncomment the lines beneath here? If this one is working as well, I'll approve and we can proceed with the merge

ozgunozerk (changes requested):

packages/tokens/non-fungible/src/test.rs
YanVictorSN: Sure. I'll dive into that.

YanVictorSN: @ozgunozerk Just did! But can't run the workflow, it's something with the `cargo-llvm-cov`

ozgunozerk: @YanVictorSN yes this bug is not related to your changes. Seems like our sccache needs to be updated. Opened a PR for that, should unblock you soon

ozgunozerk (approved):

PR #184, opened 2025-04-16
updates a broken internal cross-reference link in the nft-consecutive.adoc file

PixelPil0t1:
The original xref:tokens/non-fungible.adoc link returned a 404 error because the file path was already within the tokens directory. The corrected xref:non-fungible.adoc link uses the relative path and now works as expected in the documentation build system.


ozgunozerk:
Thanks a lot @PixelPil0t1,
but I think it is not broken. The docs are not deployed that, so one has to check whether something is broken from the `preview deploy` section in the PRs. For example, you can check that it is not broken from here: https://github.com/OpenZeppelin/stellar-contracts/pull/185#issuecomment-2808925029

It could be that, if you wanted to deploy it locally, there is a bug these intra-links are not working (they append the `[link identifier]` to the actual link, hence it is not working). Maybe that made you think it is broken (it caught me by surprise back then as well).

PR #194, opened 2025-04-23
consecutive with bitfields

### LINKED ISSUE #152: [M-03] - `owner_of` function of consecutive extension may exceed resources

brozorec:
Instead of relying on individual token IDs, this PR uses bitfields to look up ownership. It also switches from marking ownership at the first token in a batch with forward lookup to marking the last bit with backward lookup. This approach marks clearly the ownership boundaries, especially in buckets with pre-defined size.  

Fixes #152 

#### PR Checklist

- [x] Tests
- [x] Documentation


packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -369,8 +405,10 @@ impl Consecutive {
             let approval_key = StorageKey::Approval(token_id);
             e.storage().temporary().remove(&approval_key);
 
-            // Set the next token to prev owner
-            Consecutive::set_owner_for(e, from_address, token_id + 1);
+            // Set the previous token to prev owner, if there's a previous id.
+            if token_id > 0 {
+                Consecutive::set_owner_for(e, from_address, token_id - 1);
```
ozgunozerk, line 410: why don't we call:
```
e.storage().persistent().set(&StorageKey::Owner(token_id), to_address);
Self::set_ownership_in_bucket(e, token_id);
```

as in below?

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -425,6 +464,143 @@ impl Consecutive {
...
+        let mut bucket: Vec<TokenId> =
+            if let Some(b) = bucket { b } else { Vec::from_slice(e, &[0; ITEMS_IN_BUCKET]) };
+
+        // token_id in bucket
+        let relative_id = token_id - (bucket_index * ids_in_bucket);
+        let item_index = relative_id / ids_in_item;
+        let bit_index = relative_id % ids_in_item;
+
+        let mask: TokenId = 1 << (ids_in_item - bit_index - 1);
+        let mut item = bucket.get(item_index).unwrap();
```
ozgunozerk, line 505: we should replace `unwrap` with `expect` if we are sure that it won't panic with an explanation

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -425,6 +464,143 @@ impl Consecutive {
...
+        }
+
+        let bucket_index = token_id / ids_in_bucket;
+
+        let key = StorageKey::OwnershipBucket(bucket_index);
+        let bucket = e.storage().instance().get::<_, Vec<TokenId>>(&key);
+        let mut bucket: Vec<TokenId> =
+            if let Some(b) = bucket { b } else { Vec::from_slice(e, &[0; ITEMS_IN_BUCKET]) };
+
+        // token_id in bucket
```
ozgunozerk, line 499: this seemed a bit cryptic to me, maybe something more elaborate like:
`// Compute token's relative position and bit offset within its bucket` ?

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
brozorec: because we need to check the previous (`token_id - 1`) is a "free" token (not burned or not owned by someone else) and if it's the case, `set_owner_for` does exactly this

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -41,11 +43,22 @@ impl ContractOverrides for Consecutive {
     }
 }
 
-/// Storage keys for the data associated with `FungibleToken`
+/// Maximum number of vectors holding bitfields, which denote ids
```
ozgunozerk, line 46: I believe a visual demonstration via comments would be super helpful here. Something like this:

```
Bucket 2
├── Item 0 → bits for Token IDs 512..575
├── Item 1 → bits for Token IDs 576..639
├── Item 2 → bits for Token IDs 640..703
├── Item 3 → bits for Token IDs 704..767
```

If this is erroneous, feel free to fix it, or apply any other format you like :) 
This was just an example

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -381,6 +419,7 @@ impl Consecutive {
 
             // Set the new owner
             e.storage().persistent().set(&StorageKey::Owner(token_id), to_address);
+            Self::set_ownership_in_bucket(e, token_id);
```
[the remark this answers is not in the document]
brozorec: In the above case, we needed to set the previous token to preserve the ownership inference and we needed to check if it wasn't burned or owned by someone else. While here, the situation is different, we know already the current token (the token we transfer) belongs to its owner and that's why we can set it directly.

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -425,6 +464,143 @@ impl Consecutive {
...
+        let relative_id = token_id - (bucket_index * ids_in_bucket);
+        let item_index = relative_id / ids_in_item;
+        let bit_index = relative_id % ids_in_item;
+
+        let mask: TokenId = 1 << (ids_in_item - bit_index - 1);
+        let mut item = bucket.get(item_index).unwrap();
+        item |= mask;
+        // no replace, so must remove and re-insert
+        let _ = bucket.remove(item_index);
+        bucket.insert(item_index, item);
```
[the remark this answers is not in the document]
brozorec: We are using `soroban_sdk::Vec` which doesn't support indexing.

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
brozorec: like that 👌

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -425,6 +464,143 @@ impl Consecutive {
...
+    if start >= ids_in_bucket {
+        return None;
+    }
+
+    let item_index = start / ids_in_item;
+    // relative id in item
+    let relative_id = start - (item_index * ids_in_item);
+
+    (item_index..bucket.len()).find_map(|i| {
+        let from_id = if i == item_index { relative_id } else { 0 };
```
ozgunozerk, line 602: I spent 4 minutes figuring out what does this line do LOL.

If I got it right, it does:
- start from the relative index if we are inspecting the item (so we don't unnecessarily check the previous bits)
- if we are on a later item, then we have to start from `0`

if my understanding is correct, can we put some inline docs here? This whole bitmap thing is probably where our library is working at the most low-level, so I think it deserves some extra inline comments

ozgunozerk (changes requested):
Amazing! I always love me some bitmaps ❤️ 
Few nitpicks, but great in general 💯

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
ozgunozerk: ok great, can we also document this as inline comments? If I couldn't understand it, probably other people will have similar problems

packages/tokens/non-fungible/src/extensions/consecutive/storage.rs
```diff
@@ -425,6 +464,143 @@ impl Consecutive {
...
+///
+/// # Example
+///
+/// If `IDS_IN_ITEM = 8`, `ITEMS_IN_BUCKET = 2`, and:
+/// ```
+/// bucket = vec![0b00000000, 0b00101000];
+/// ```
+/// then `find_bit_in_bucket(bucket, 8)` returns `Some(11)`, since bit 3 of the
+/// second item (index 11 in the overall bucket) is the first set bit.
+pub(crate) fn find_bit_in_bucket(bucket: Vec<TokenId>, start: TokenId) -> Option<TokenId> {
```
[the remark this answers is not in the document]
brozorec: Tried by using refs but it requires even more back-and-forth casting between `usize` and `u32` that's used by  `soroban_sdk::Vec`.

ozgunozerk (approved):

PR #215, opened 2025-05-07
SacAdmin

### LINKED ISSUE #207: SAC Admin
The current `token::StellarAssetClient` in "soroban_sdk" provides built-in support for managing SAC (Stellar Asset Contract) administrative tasks. However, some use cases require more flexible or customized logic.

To address this, we will implement a dedicated SAC Admin module to which the admin tasks can be delegated to and that replicates the admin functionality of `token::StellarAssetClient`. This allows developers to define custom rules (e.g., role-based access, multi-step admin transfers, mint rate limits) while maintaining compatibility with SACs.

For more context, see:
https://developers.stellar.org/docs/build/guides/tokens/custom-sac-admin

brozorec:
Fixes #207 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/tokens/fungible/src/utils/sac_admin/mod.rs
+++ b/packages/tokens/fungible/src/utils/sac_admin/mod.rs
@@ -1,0 +1,25 @@
+mod storage;
+
+mod test;
+
+use soroban_sdk::{Address, Env};
+
+/// Mintable Trait for Fungible Token
+///
+/// The `FungibleMintable` trait extends the `FungibleToken` trait to provide
+/// the capability to mint tokens. This trait is designed to be used in
+/// conjunction with the `FungibleToken` trait.
+///
+/// Excluding the `mint` functionality from the
+/// [`crate::fungible::FungibleToken`] trait is a deliberate design choice to
+/// accommodate flexibility and customization for various smart contract use
+/// cases.
+pub trait SACAdmin {
+    fn mint(e: Env, to: Address, amount: i128, operator: Address);
+
+    fn set_admin(e: Env, new_admin: Address, operator: Address);
+
+    fn set_authorized(e: Env, id: Address, authorize: bool, operator: Address);
+
+    fn clawback(e: Env, from: Address, amount: i128, operator: Address);
+}
--- a/packages/tokens/fungible/src/utils/sac_admin/mod.rs
+++ b/packages/tokens/fungible/src/utils/sac_admin/mod.rs
@@ -1,19 +1,9 @@
-mod storage;
+pub mod storage;
 
 mod test;
 
 use soroban_sdk::{Address, Env};
 
-/// Mintable Trait for Fungible Token
-///
-/// The `FungibleMintable` trait extends the `FungibleToken` trait to provide
-/// the capability to mint tokens. This trait is designed to be used in
-/// conjunction with the `FungibleToken` trait.
-///
-/// Excluding the `mint` functionality from the
-/// [`crate::fungible::FungibleToken`] trait is a deliberate design choice to
-/// accommodate flexibility and customization for various smart contract use
-/// cases.
 pub trait SACAdmin {
     fn mint(e: Env, to: Address, amount: i128, operator: Address);
 
--- a/packages/tokens/fungible/src/utils/sac_admin/storage.rs
+++ b/packages/tokens/fungible/src/utils/sac_admin/storage.rs
@@ -1,0 +1,44 @@
+use soroban_sdk::{contracttype, panic_with_error, token::StellarAssetClient, Address, Env};
+
+use crate::FungibleTokenError;
+
+#[contracttype]
+pub enum SacDataKey {
+    Sac,
+}
+
+pub fn set_sac_address(e: &Env, sac: &Address) {
+    e.storage().instance().set(&SacDataKey::Sac, &sac);
+}
+
+pub fn get_sac_address(e: &Env) -> Address {
+    e.storage()
+        .instance()
+        .get(&SacDataKey::Sac)
+        .unwrap_or_else(|| panic_with_error!(e, FungibleTokenError::SACNotSet))
+}
+
+pub fn get_sac_client(e: &Env) -> StellarAssetClient {
+    let sac_address = get_sac_address(e);
+    StellarAssetClient::new(e, &sac_address)
+}
+
+pub fn mint(e: &Env, to: &Address, amount: i128) {
+    let client = get_sac_client(e);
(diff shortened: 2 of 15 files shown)


packages/tokens/fungible/src/utils/sac_admin/storage.rs
```diff
@@ -0,0 +1,121 @@
...
+
+// ################## QUERY STATE ##################
+
+/// Stores the SAC address, typically called from the constructor.
+///
+/// # Arguments
+///
+/// * `e` - Access to Soroban environment.
+/// * `sac` - The address the SAC contract.
+pub fn set_sac_address(e: &Env, sac: &Address) {
```
ozgunozerk, line 19: in some of our (and/or OpenZeppelin's contracts') other designs (the most recent one is `ownable`), we don't have a `setter` for the things that are meant to be initialized at constructor. 

I'm not against this, but if we are going to follow having a separate `setter` function, we should do that consistently across our library.

Just a discussion point 👍

packages/tokens/fungible/src/utils/sac_admin/test.rs
```diff
@@ -0,0 +1,130 @@
...
+    contract,
+    testutils::{Address as _, MockAuth, MockAuthInvoke, StellarAssetContract},
+    token::{StellarAssetClient, TokenClient},
+    xdr::AccountFlags,
+    Address, Env, IntoVal,
+};
+use soroban_test_helpers;
+
+use super::storage::{clawback, set_admin, set_authorized};
+use crate::sac_admin::storage::{get_sac_address, mint, set_sac_address};
```
ozgunozerk, line 15: aren't those two are the same?

Btw, the keyword `super` is prone to misunderstandings. I'd advocate for using `crate::FULL_PATH_HERE` instead of `super` in general

packages/tokens/fungible/src/utils/sac_admin/test.rs
```diff
@@ -0,0 +1,130 @@
...
+    let token_client = TokenClient::new(&e, &sac_client.address);
+    e.as_contract(&new_admin, || {
+        mint(&e, &user, 100);
+        assert_eq!(token_client.balance(&user), 100);
+    });
+}
+
+#[soroban_test_helpers::test]
+#[should_panic(expected = "Error(Contract, #11)")]
+// topics:[error, Error(Contract, #11)], data:"balance is deauthorized"
```
ozgunozerk, line 91: should we remove this one?

ozgunozerk (changes requested):
Looking solid!

There is one point that I'm confused on, so maybe we can improve the inline docs to eliminate this confusion for others as well.

In almost all of our other crates' functions, we have `caller` parameter. We don't in this one.

For example:
```
pub fn set_admin(e: &Env, new_admin: &Address) {
    let client = get_sac_client(e);
    client.set_admin(new_admin);
}
```

However, in the trait method, we do have an `operator`:
```
fn set_admin(e: Env, new_admin: Address, operator: Address);
```

So, why we don't have `operator` in the storage functions? 

From the tests, I saw that `operator` is not needed, and one can call the `set_admin` function like this:
```
    e.as_contract(new_admin, || {
        set_sac_address(e, sac);
        set_admin(e, new_admin);
    });
```

Then, why do we have it on trait? 

Clarifications on these could be really helpful imo.

That's my only concern about it, otherwise, 💯

packages/tokens/fungible/src/utils/sac_admin/test.rs
brozorec: yes, it's coming from the sdk, will doc it better

packages/tokens/fungible/src/utils/sac_admin/storage.rs
brozorec: Provided some reasoning about that in the AC PR and I think you're also in favor for having explicit setters.

brozorec:
@ozgunozerk I provided some explanations about the point you mentioned. 
I also have a question: Do you think the crate and the trait names are fine, given we'll have to implement the other approach in the next milestone? Will it be better to rename the current one to `sac_admin_wrapper` straightaway?

ozgunozerk:
> Do you think the crate and the trait names are fine, given we'll have to implement the other approach in the next milestone? Will it be better to rename the current one to sac_admin_wrapper straightaway?

Yes, i'd vote for having the final version of the name from the beginning since the trajectory is almost certain in this case 👍

packages/tokens/fungible/src/utils/sac_admin/README.md
```diff
@@ -0,0 +1,41 @@
+```mermaid
+sequenceDiagram
+    actor Issuer
+    actor Owner
+    participant Admin
+    participant AdminSigner
```
ozgunozerk, line 6: I think a few explanations on top of this mermaid diagram would be useful. For example:
- `Admin` is the wrapper contract
- `Owner` can be renamed (may be confusing due to we have ownable etc.)
- an intro like: "this contract is a wrapper contract that can be deployed on top of an existing contract to delegate admin related functions, etc.

packages/tokens/fungible/src/utils/sac_admin/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+pub use storage::{
+    clawback, get_sac_address, get_sac_client, mint, set_admin, set_authorized, set_sac_address,
+};
+
+mod test;
+
+use soroban_sdk::{Address, Env};
+
+pub trait SACAdmin {
+    /// Sets the administrator to the specified address `new_admin`.
```
ozgunozerk, line 70: Would you agree there is value in specifying `administrator contract` instead of `administrator`? I believe this will help people understand the concept better. Wdyt?

ozgunozerk (approved):
LGTM!

I think there are few points to improve the docs (I've commented them already), aside from that, 💯

ozgunozerk:
The folder name should be `sac-admin` instead of `sac_admin` under `packages/tokens/fungible/src/utils/sac_admin`

packages/tokens/fungible/src/utils/sac_admin/mod.rs
brozorec: They can set a G-account as a new administrator, it's not required to be a contract.

brozorec:
> The folder name should be `sac-admin` instead of `sac_admin` under `packages/tokens/fungible/src/utils/sac_admin`

@ozgunozerk This folder is a mod, not a crate and that's why using `_`.

Issue #209, opened 2025-04-29
Ownable with 2-Step Ownership Transfer


brozorec:
This module extends the basic `ownable`  by introducing a 2-step ownership transfer process. It is designed to reduce the risk of accidental transfers by requiring the new owner to explicitly accept the role.

Ownership transfer begins with a call to `transfer_ownership`, which nominates a new owner. The transfer is only finalized when the nominated owner calls `accept_ownership`.

Do we include in the scope a configurable delay between the two steps, with the ability to cancel before the transfer gets accepted?

PR #216, opened 2025-05-10
Ownable

### LINKED ISSUE #208: Ownable
Implement a simple access control module that defines an "owner" account with exclusive permission to execute specific contract functions. This is useful for contracts where a single authority is responsible for sensitive operations (e.g., configuration, upgrades, or privileged actions).

The initial owner is set at deployment using an address provided by the deployer. Ownership can be transferred using a `transfer_ownership` function or renounced entirely with `renounce_ownership`.

Events:
- `ownership_transferred`
- `ownership_renounced`

To make it feel as a function modifier, the module will also include a procedural macro attribute `#[only_owner]`, which can be applied to contract functions to restrict their execution to the current owner (if owner has renounced, function can be called anymore). The macro will expand to include `ensure_is_owner()` that checks the authorization of the owner for this invocation.

To further reduce boilerplate, do we include a derive macro that provides default implementations for the functions from this module? Can the defaults be included in `#[default_impl]`?

ozgunozerk:
Fixes #208 and #209 

#### PR Checklist

- [x] Tests
- [x] Documentation

This turned out to be a compound PR. Code-wise, it may seem big, but aside from `ownable`, it's pretty much all refactoring related things.

Here is a breakdown:
- implements `ownable`
- implements `#[only_owner]` macro in `ownable-macro` crate
- while implementing the `#[only_owner]` macro, I realized there are some re-usable parts across some of our macros. So decided to have a helper macro crate called `stellar-macro-helpers`. 
- refactored `pausable-macros` accordingly due to this new `stellar-macro-helpers` crate.
- according to the suggestion, made `2-step-role-transfer` re-usable by exporting it into another crate, that can be shared across access control and ownable. And it uses generics :) 
- now, we have several helper crates (`role-transfer`, `macro-helpers`), I suggest we create yet another folder, called `helpers`, and we transfer all these crates which are not meant to be used directly by end-developers. So that, `contract-utils` only consists of crates which are meant to be directly used by end-developers. If agreed, I'll make that change as well @brozorec 
- `#[default_impl]` support for `Ownable` is there. We don't have `#[default_impl]` for `Pausable`, should we reconsider it, or should we deliberately leave that one up to the end-developer to implement it? This may be a good discussion for later (does not concern this PR).
- more readable format for codecov cli command in github workflow file.

Note: draft because should be merged after `access-control`


ozgunozerk:
https://github.com/OpenZeppelin/stellar-contracts/pull/214#discussion_r2084674948

should be resolved before merging this

brozorec (changes requested):
>now, we have several helper crates (role-transfer, macro-helpers), I suggest we create yet another folder, called helpers, and we transfer all these crates which are not meant to be used directly by end-developers. So that, contract-utils only consists of crates which are meant to be directly used by end-developers. If agreed, I'll make that change as well @brozorec

Sounds good 👍 

---

Conceptually looks good overall, great job! Beside the comment about `#[only_owner]`, there are some nitpicks, but we'll clear them later. Ok for you?

packages/contract-utils/ownable-macro/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+#[proc_macro_attribute]
+pub fn only_owner(_attrs: TokenStream, input: TokenStream) -> TokenStream {
+    let input_fn = parse_macro_input!(input as ItemFn);
+
+    // Use the utility function to get the environment parameter
+    let env_param = parse_env_arg(&input_fn);
+
+    // Find the first Address parameter
+    let (address_param, is_ref) = find_address_param(&input_fn)
+        .expect("No parameter of type Address or &Address found in function signature");
```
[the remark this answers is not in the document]
ozgunozerk: I think you misinterpreted the code. The macro does not require an address to be passed to it. 

It tries to find the first `Account` types parameter in the function it is annotating to enforce the check.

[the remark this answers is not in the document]
brozorec: I didn't misinterpreted it, just poorly explained my point, sorry. I meant the function on which the macro gets applied doesn't need to have an account passed as parameter.

Let me illustrate it with examples. Let's say we want to apply the macro on:
```
    pub fn increment(e: &Env) -> i32 {
        let owner: Address = e.storage().instance().get(&DataKey::Owner).unwrap();
        owner.require_auth();
        // increment counter
    }
```
so that it looks like:
```
    #[only_owner]
    pub fn increment(e: &Env) -> i32 {
        // increment counter
    }
```
We don't need to pass `caller: Address` in the first snippet to check the authorization, so the macro doesn't need to look for such a function param in order to be able to do its job

packages/contract-utils/role-transfer/src/storage.rs
```diff
@@ -0,0 +1,78 @@
...
+use crate::RoleTransferError;
+
+/// Initiates the role transfer (owner/admin).
+/// If `live_until_ledger == 0`, cancels the pending transfer.
+///
+/// Does not emit any events.
+///
+/// Returns None if the transfer was initiated, Some(Address) where the Address
+/// is the new candidate for the pending role in the storage if the pending
+/// transfer is revoked.
```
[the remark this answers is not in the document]
brozorec: Also add for both functions the arguments and the errors lists in the annotations, although they are probably meant to be used only internally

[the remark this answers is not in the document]
brozorec: Btw I find it a bit confusing to return `Some` when removing a key and `None` when setting. I'd expect this to happen the other way around. Why do we need to return whatsoever? Is it because of the need to query the pending one and remove it in one shot?

packages/contract-utils/role-transfer/src/storage.rs
```diff
@@ -0,0 +1,78 @@
...
+where
+    T: IntoVal<Env, Val>,
+    U: IntoVal<Env, Val>,
+{
+    current.require_auth();
+
+    let current_in_storage = e
+        .storage()
+        .instance()
+        .get::<_, Address>(active_key)
```
brozorec, line 30:
```suggestion
        .get::<T, Address>(active_key)
```
so that we use it explicitly :) feel free to ignore it

packages/contract-utils/role-transfer/src/storage.rs
```diff
@@ -0,0 +1,78 @@
...
+        .unwrap_or_else(|| panic_with_error!(e, RoleTransferError::AccountNotFound));
+
+    if current != &current_in_storage {
+        panic_with_error!(e, RoleTransferError::Unauthorized);
+    }
+
+    if live_until_ledger == 0 {
+        let Some(pending) = e.storage().temporary().get::<_, Address>(pending_key) else {
```
brozorec, line 38:
```suggestion
        let Some(pending) = e.storage().temporary().get::<U, Address>(pending_key) else {
```

packages/contract-utils/ownable/src/ownable.rs
```diff
@@ -0,0 +1,112 @@
...
+    ///   owner.
+    fn renounce_ownership(e: &Env, caller: Address);
+}
+
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum OwnableError {
+    NotAuthorized = 200,
+    CannotRenounceWhilePendingTransfer = 201,
```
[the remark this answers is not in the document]
ozgunozerk: Shortened the name with `TransferInProgress`. I think having 2 different errors here make more sense.
Resolving this one for now since error codes are also fixed.
Feel free to unresolve this one if you think otherwise

packages/contract-utils/role-transfer/src/storage.rs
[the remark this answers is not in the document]
ozgunozerk: I reworked it, now it doesn't return anything. Also addressed the other points.

packages/contract-utils/ownable/src/storage.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// * See [`accept_transfer()`] for possible errors.
+///
+/// # Events
+///
+/// * topics - `["ownership_transfer_completed"]`
+/// * data - `[old_owner: Address, new_owner: Address]`
+pub fn accept_ownership(e: &Env, caller: &Address) {
+    let previous_owner =
+        get_owner(e).expect("One cannot renounce ownership while there is a pending transfer.");
```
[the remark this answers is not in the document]
ozgunozerk: You are right. The more I thought on it, the more unnecessary it started to seem to me.

By it, I mean, having `old_owner` in the `transfer_complete` event.

The whole point of querying for `previous_owner` here, is to have that data emitted in the event. However, this data (`old_owner`) is already emitted with `transfer_initiated` event. No need to duplicate this information for sequential events imo. 

So, I took the liberty and removed `old_owner` from that event altogether. It is still open for discussion and I can rewind the changes.

Coming to your suggestion here, if we decide to roll-back the changes I've made -> then I agree your suggestions are on point, and I'll implement them 👍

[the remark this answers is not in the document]
brozorec: > The whole point of querying for previous_owner here, is to have that data emitted in the event. However, this data (old_owner) is already emitted with transfer_initiated event. No need to duplicate this information for sequential events imo.

That makes sense, I agree 👍

packages/contract-utils/access-control/src/access_control.rs
```diff
@@ -131,25 +133,31 @@ pub trait AccessControl {
...
     ///   cancelled. `live_until_ledger` argument is implicitly bounded by the
     ///   maximum allowed TTL extension for a temporary storage entry and
     ///   specifying a higher value will cause the code to panic.
     ///
-    /// # Errors
-    ///
-    /// * `AccessControlError::Unauthorized` - If the `caller` is not the admin.
-    /// * `AccessControlError::NoPendingAdminTransfer` - If tried to cancel the
-    ///   pending admin transfer when there is no pending admin transfer.
+    /// * [`stellar_role_transfer::RoleTransferError::NoPendingTransfer`] - If
```
brozorec, line 143: missing `# Errors`

Not a fan of listing those errors from another crate in this trait, although I don't have a better idea how to handle this, given the current structure.

packages/contract-utils/access-control/src/storage.rs
```diff
@@ -399,8 +357,9 @@ pub fn set_role_admin(e: &Env, admin: &Address, role: &Symbol, admin_role: &Symb
 ///
 /// # Errors
 ///
-/// * `AccessControlError::Unauthorized` - If the caller is neither the contract
-///   admin nor has the admin role.
+/// * refer to [`get_admin`] errors.
+/// * [`AccessControlError::Unauthorized`] - If the caller is neither the
+///   contract admin nor has the admin role.
```
brozorec, line 362:
```suggestion
/// * [`AccessControlError::Unauthorized`] - If the caller is neither the
///   contract admin nor has the admin role.
/// * refer to [`get_admin`] errors.
```

packages/contract-utils/ownable/src/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+#[contracttype]
+pub enum OwnableStorageKey {
+    Owner,
+    PendingOwner,
+}
+
+// ################## QUERY STATE ##################
+
+/// Returns `Some(Address)` if ownership is set, or `None` if ownership has been
+/// renounced.
```
brozorec, line 19:
```suggestion
/// renounced or has never been set.
```

packages/contract-utils/ownable/src/ownable.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// * [`stellar_role_transfer::RoleTransferError::NoPendingTransfer`] - If
+    ///   there is no pending transfer to accept.
+    /// * [`stellar_role_transfer::RoleTransferError::Unauthorized`] - If the
+    ///   caller is not the pending role holder.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["ownership_transfer_completed"]`
+    /// * data - `[new_owner: Address]`
+    fn accept_ownership(e: &Env, caller: Address);
```
brozorec, line 65: Do we actually need the `caller` param here? It is supposed to be in the storage and can be fetched from there.

If you agree that we don't need it, that will affect the storage function in this module and `stellar_role_transfer::accept_transfer`.

packages/contract-utils/ownable/src/storage.rs
```diff
@@ -0,0 +1,153 @@
...
+///
+/// # Errors
+///
+/// * refer to [`accept_transfer`] errors.
+///
+/// # Events
+///
+/// * topics - `["ownership_transfer_completed"]`
+/// * data - `[new_owner: Address]`
+pub fn accept_ownership(e: &Env, caller: &Address) {
```
brozorec, line 90:
```suggestion
/// * data - `[new_owner: Address]`
///
/// # Notes
///
/// * Authorization for the pending owner is required.
pub fn accept_ownership(e: &Env, caller: &Address) {
```

brozorec (changes requested):

packages/contract-utils/access-control/src/access_control.rs
ozgunozerk: I have the EXACT same feelings :D

packages/contract-utils/ownable/src/ownable.rs
ozgunozerk: good catch, also changed the `accept_admin`

brozorec (approved):
LGTM 👌

PR #221, opened 2025-05-19
Royalty

### LINKED ISSUE #220: 🏗️ [Core Feature]: Implement Royalty extension for Non Fungible Tokens

Following ERC2981, we should provide Royalty functionality for Non Fungible Tokens

ozgunozerk:
Fixes #220 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
+++ b/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
@@ -78,17 +78,13 @@ pub trait NonFungibleRoyalties: NonFungibleToken {
     ///   for a token that already has royalty information.
     fn set_token_royalty(e: &Env, token_id: u32, receiver: Address, basis_points: u32);
 
-    /// Returns the royalty information for a token.
+    /// Returns `(Address, u32)` - A tuple containing the receiver address and the royalty amount.
     ///
     /// # Arguments
     ///
     /// * `e` - Access to the Soroban environment.
     /// * `token_id` - The identifier of the token.
     /// * `sale_price` - The sale price for which royalties are being calculated.
-    ///
-    /// # Returns
-    ///
-    /// * `(Address, u32)` - A tuple containing the receiver address and the royalty amount.
     ///
     /// # Errors
     ///
--- a/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
+++ b/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
@@ -1,0 +1,97 @@
+mod storage;
+use crate::NonFungibleToken;
+
+mod test;
+
+use soroban_sdk::{Address, Env};
+
+/// Maximum allowed royalty percentage (5000 basis points = 50%)
+pub const MAX_ROYALTY_BASIS_POINTS: u32 = 5000;
+
+/// Royalties Trait for Non-Fungible Token (ERC2981)
+///
+/// The `NonFungibleRoyalties` trait extends the `NonFungibleToken` trait to
+/// provide the capability to set and query royalty information for tokens. This trait
+/// is designed to be used in conjunction with the `NonFungibleToken` trait.
+///
+/// This implementation follows the ERC2981 standard for royalties, allowing:
+/// - Setting global royalties for the entire collection
+/// - Setting per-token royalties that override the global setting
+/// - Enforcing a maximum royalty percentage to prevent scams
+/// - Making royalties immutable after minting
+///
+/// `storage.rs` file of this module provides the `NonFungibleRoyalties` trait
+/// implementation for the `Base` contract type. For other contract types (eg.
+/// `Enumerable`, `Consecutive`), the overrides of the `NonFungibleRoyalties`
+/// trait methods can be found in their respective `storage.rs` file.
+///
+/// # Notes
+///
+/// `#[contractimpl]` macro requires even the default implementations to be
+/// present under its scope. To not confuse the developers, we did not provide
+/// the default implementations here, but we are providing a macro to generate
+/// the default implementations for you.
+///
+/// When implementing [`NonFungibleRoyalties`] trait for your Smart Contract,
+/// you can follow the below example:
+///
+/// ```ignore
+/// #[default_impl] // **IMPORTANT**: place this above `#[contractimpl]`
+/// #[contractimpl]
+/// impl NonFungibleRoyalties for MyContract {
+///     /* your overrides here (you don't have to put anything here if you don't want to override anything) */
+///     /* and the macro will generate all the missing default implementations for you */
+/// }
+/// ```
+pub trait NonFungibleRoyalties: NonFungibleToken {
+    /// Sets the global default royalty information for the entire collection.
+    /// This will be used for all tokens that don't have specific royalty information.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `receiver` - The address that should receive royalty payments.
+    /// * `basis_points` - The royalty percentage in basis points (100 = 1%, 10000 = 100%).
+    ///
+    /// # Errors
+    ///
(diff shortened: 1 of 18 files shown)


janewang:
> there is a cap for setting the royalty amount. I've made it 50% (questionable...). I'm quite indecisive about whether to make this tweak-able by exposing a setter function. Currently, it makes more sense to me to have it as a global constant. Because the main point of having this cap is to again prevent scams. If we allow people to tweak this, then, idk...

Why do we need a max? I don't think this requirement is necessary and 50% is definitely too low for some cases.

packages/tokens/non-fungible/src/extensions/royalties/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+/// This implementation follows the ERC2981 standard for royalties, allowing:
+/// - Setting global royalties for the entire collection
+/// - Setting per-token royalties that override the global setting
+/// - Enforcing a maximum royalty percentage to prevent scams
+/// - Making royalties immutable after minting
+///
+/// `storage.rs` file of this module provides the `NonFungibleRoyalties` trait
+/// implementation for the `Base` contract type. For other contract types (eg.
+/// `Enumerable`, `Consecutive`), the overrides of the `NonFungibleRoyalties`
+/// trait methods can be found in their respective `storage.rs` file.
```
brozorec, line 24: I don't see the overrides mentioned for `Enumerable` and `Consecutive`.

packages/tokens/non-fungible/src/extensions/royalties/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+///
+/// # Notes
+///
+/// `#[contractimpl]` macro requires even the default implementations to be
+/// present under its scope. To not confuse the developers, we did not provide
+/// the default implementations here, but we are providing a macro to generate
+/// the default implementations for you.
```
brozorec, line 31:
```suggestion
/// present under its scope. To avoid confusion, we do not provide the default
/// implementations here, but we are providing a macro that generates them
///  for you.
```

packages/tokens/non-fungible/src/extensions/royalties/storage.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Storage container for royalty information
+#[contracttype]
+pub struct RoyaltyInfo {
+    pub receiver: Address,
+    pub basis_points: u32,
+}
+
+/// Storage keys for royalty data
+#[contracttype]
+pub enum RoyaltyStorageKey {
```
brozorec, line 15: rename to `NFTRoyaltiesStorageKey` for consistency with the other extensions.

packages/tokens/non-fungible/src/extensions/royalties/storage.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///   10000 = 100%).
+    ///
+    /// # Errors
+    ///
+    /// * [`NonFungibleTokenError::RoyaltyTooHigh`] - If the royalty percentage
+    ///   exceeds the maximum allowed value.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["royalty", receiver: Address]`
```
brozorec, line 39: - this event is not emitted
- I'd suggest renaming the 1st topic to `default_royalty` or `collection_royalty`
- Should we exclude `receiver` from topics to be consistent with the event below (or the other way around)

packages/tokens/non-fungible/src/extensions/royalties/storage.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///   exist.
+    /// * [`NonFungibleTokenError::RoyaltyTooHigh`] - If the royalty percentage
+    ///   exceeds the maximum allowed value.
+    /// * [`NonFungibleTokenError::RoyaltyAlreadySet`] - If attempting to set
+    ///   royalties for a token that already has royalty information.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["tokroyal", token_id: u32]`
+    /// * data - `[receiver: Address, basis_points: u32]`
```
brozorec, line 78: - this event is not emitted
- I'd suggest renaming the 1st topic to `token_royalty`

packages/tokens/non-fungible/src/extensions/royalties/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+///
+/// The `NonFungibleRoyalties` trait extends the `NonFungibleToken` trait to
+/// provide the capability to set and query royalty information for tokens. This
+/// trait is designed to be used in conjunction with the `NonFungibleToken`
+/// trait.
+///
+/// This implementation follows the ERC2981 standard for royalties, allowing:
+/// - Setting global royalties for the entire collection
+/// - Setting per-token royalties that override the global setting
+/// - Enforcing a maximum royalty percentage to prevent scams
```
brozorec, line 18: Is this still the case?

packages/tokens/non-fungible/src/extensions/royalties/storage.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `token_id` - The identifier of the token.
+    /// * `sale_price` - The sale price for which royalties are being
+    ///   calculated.
+    ///
+    /// # Returns
+    ///
+    /// * `(Address, u32)` - A tuple containing the receiver address and the
```
brozorec, line 112: mention that if no token or default royalty, it returns the contract address

brozorec (changes requested):
- the param`basis_points` should be restricted to 10_000 and panic if the value is above when trying to set it
- `NonFungibleTokenError::RoyaltyTooHigh` is no longer used

janewang:
> once a royalty is set for a token, it cannot be set again (to prevent scams)

I think this is questionable. Maybe should be based on implementation.

packages/tokens/non-fungible/src/extensions/royalties/mod.rs
ozgunozerk: Thanks! And let's not resolve this convo as a reminder, maybe the discussion with Jane result in another change 👍

janewang:
> Could you elaborate on this?

> I know that it is a restriction, but to my understanding, changing the royalty for an NFT that has been already published has little to none actual use-case. On the other hand, allowing this functionality can indeed be dangerous.


@ozgunozerk  Royalty is closely tied to a project's business model, therefore flexibility is needed to modify the business based on market demand. For example, project may start with a certain royalty, but later partners, investors, and contributors may require a different revenue split as project matures. There are stakeholders whose needs evolve as a business grows or declines and the split will require to change. I understand the concern for safety but it should be based on implementation.

ozgunozerk:
@janewang makes sense! 
Will update the code accordingly 🚀

ozgunozerk:
3rd suggestion actually makes sense 👍 

1st and 2nd one are not though :)

packages/tokens/non-fungible/src/extensions/royalties/storage.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        // Verify token exists by checking owner
+        let _ = Base::owner_of(e, token_id);
+
+        // Store the token royalty information
+        let key = NFTRoyaltiesStorageKey::TokenRoyalty(token_id);
+        let royalty_info = RoyaltyInfo { receiver: receiver.clone(), basis_points };
+        e.storage().persistent().set(&key, &royalty_info);
+        e.storage().persistent().extend_ttl(&key, OWNER_TTL_THRESHOLD, OWNER_EXTEND_AMOUNT);
```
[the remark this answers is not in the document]
ozgunozerk: good catch, also fixed AC module

PR #237, opened 2025-05-30
allowlist and blocklist

### LINKED ISSUE #236: 🏗️ [Core Feature]: Allowlist and Blocklist for Fungible Token

Similar to Ethereum, we can implement Allowlist and Blocklist for Stellar Soroban Smart Contracts

ozgunozerk:
Fixes #236

#### PR Checklist

- [x] Tests
- [x] Documentation

Draft, because `fungible-refactor` should merge first.


ozgunozerk:
Will add tests for coverage soon

brozorec (changes requested):
I'd strongly advocate to not include the admin handling in the two new extensions. We already have `ownable` (with 2-step transfer) and `access_control` (with 2-step admin transfer) that are a great fit for such a kind of cases. Baking-in proper basic admin functionality in those extensions is limiting for slightly more complex real-world cases and feels like a less flexible version of existing features.
I suggest to not manage authorizations in allow/disallow and block/unblock functions by marking them as low-level functions that are meant to be embedded in other higher-level functions. Furthermore, if we want to keep the traits `FungibleBlockList` and `FungibleAllowList`, we might want to add an additional `operator: Address` param to the setter functions, which will allow checking authorizations and permissions when implementing them.

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Allows a user to receive and transfer tokens.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `user` - The address to allow.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["user_allowed", user: Address]`
```
brozorec, line 58:
```suggestion
    /// * topics - `["allow", user: Address]`
```

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Disallows a user from receiving and transferring tokens.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `user` - The address to disallow.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["user_disallowed", user: Address]`
```
brozorec, line 90:
```suggestion
    /// * topics - `["disallow", user: Address]`
```

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// - In admin functions that implement their own authorization logic
+    ///
+    /// Using this function in public-facing methods creates significant
+    /// security risks as it could allow unauthorized allowlist
+    /// modifications.
+    pub fn allow_user(e: &Env, user: &Address) {
+        // Set the user as allowed
+        let key = AllowListStorageKey::Allowed(user.clone());
+        e.storage().persistent().set(&key, &true);
+        e.storage().persistent().extend_ttl(&key, BALANCE_TTL_THRESHOLD, BALANCE_EXTEND_AMOUNT);
```
brozorec, line 75: we don't need to extend persistent storage when it gets set, only when it's read

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    /// Returns the allowed status of an account.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `account` - The address to check the allowed status for.
+    pub fn allowed(e: &Env, account: &Address) -> bool {
+        let key = AllowListStorageKey::Allowed(account.clone());
+        e.storage().persistent().get(&key).unwrap_or(false)
```
brozorec, line 44: extend ttl

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    ///   not allowed.
+    /// * Also refer to [`Base::transfer`] errors.
+    pub fn transfer(e: &Env, from: &Address, to: &Address, amount: i128) {
+        // Check if both addresses are allowed
+        if !AllowList::allowed(e, from) {
+            panic_with_error!(e, FungibleTokenError::UserNotAllowed);
+        }
+        if !AllowList::allowed(e, to) {
+            panic_with_error!(e, FungibleTokenError::UserNotAllowed);
+        }
```
brozorec, line 136:
```suggestion
        if !AllowList::allowed(e, from) || !AllowList::allowed(e, to) {
            panic_with_error!(e, FungibleTokenError::UserNotAllowed);
        }
```

packages/tokens/fungible/src/extensions/allowlist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// # Errors
+    ///
+    /// * [`FungibleTokenError::UserNotAllowed`] - When either `from`, `to`, or
+    ///   `spender` is not allowed.
+    /// * Also refer to [`Base::transfer_from`] errors.
+    pub fn transfer_from(e: &Env, spender: &Address, from: &Address, to: &Address, amount: i128) {
+        // Check if all addresses are allowed
+        if !AllowList::allowed(e, spender) {
+            panic_with_error!(e, FungibleTokenError::UserNotAllowed);
+        }
```
brozorec, line 164: this case is an interesting one, do we need to check `spenser`, because technically they don't operate with their own funds

packages/tokens/fungible/src/extensions/allowlist/storage.rs
brozorec: seeing this as pattern in all functions below, is there a particular reason to list them one by one?

packages/tokens/fungible/src/extensions/blocklist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Blocks a user from receiving and transferring tokens.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `user` - The address to block.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["user_blocked", user: Address]`
```
brozorec, line 58:
```suggestion
    /// * topics - `["block", user: Address]`
```

packages/tokens/fungible/src/extensions/blocklist/storage.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Unblocks a user, allowing them to receive and transfer tokens.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `user` - The address to unblock.
+    ///
+    /// # Events
+    ///
+    /// * topics - `["user_unblocked", user: Address]`
```
brozorec, line 90:
```suggestion
    /// * topics - `["unblock", user: Address]`
```

packages/tokens/fungible/src/extensions/allowlist/storage.rs
ozgunozerk: yes, just double checked with solidity library, they didn't put such a bound, removing this one

brozorec (approved):
LGTM 👍 

One last note:
The query functions `allowed(e: &Env, account: &Address)` and `blocked(e: &Env, account: &Address)` use `account` instead of `user` (as it's in all other functions, traits and events). This inconsistent usage is also present in the sol lib. I'd prefer to have consistent naming across both module, but if you think it's better to leave it that way, I'm fine too.

PR #252, opened 2025-06-11
default impl support for royalties

### LINKED ISSUE #249: Default impl macro support for royalties
Also, adding inline documentation about why `allowlist` and `blocklist` are not supported by default impl macro would be useful

ozgunozerk:
Fixes #249 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
+++ b/packages/tokens/non-fungible/src/extensions/royalties/mod.rs
@@ -59,7 +59,6 @@ pub trait NonFungibleRoyalties: NonFungibleToken {
     /// * data - `[basis_points: u32]`
     fn set_default_royalty(e: &Env, receiver: Address, basis_points: u32, operator: Address);
 
-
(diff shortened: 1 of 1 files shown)


packages/contract-utils/default-impl-macro/src/lib.rs
```diff
@@ -47,8 +47,20 @@ mod helper;
...
+/// # Notes
+///
+/// This macro does not support the below traits on purpose:
+/// - `FungibleAllowList`
+/// - `FungibleBlockList`
+///
+/// Because, there are no default implementation to enforce how the
+/// authorization should be configured. Not providing a default implementation
+/// for these traits is basically acting as a reminder for the implementor to
+/// provide the authorization logic for these traits.
```
brozorec, line 63: Doesn't this argument apply to `NonFungibleRoyalties` as well? Because both setters are unauthorized functions.

PR #297, opened 2025-06-25
SAC admin security warnings

### LINKED ISSUE #283: [N-06] Missing Warning When Changing SAC Admin Address

brozorec:
Fixes #283 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/tokens/fungible/src/utils/sac_admin_generic/mod.rs
+++ b/packages/tokens/fungible/src/utils/sac_admin_generic/mod.rs
@@ -40,6 +40,12 @@
 //! - The wrapper approach is simpler to implement and more flexible but
 //!   requires additional entry points for each admin function.
 //!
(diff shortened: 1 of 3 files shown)


packages/tokens/fungible/src/utils/sac_admin_generic/mod.rs
```diff
@@ -40,6 +40,12 @@
 //! - The wrapper approach is simpler to implement and more flexible but
 //!   requires additional entry points for each admin function.
 //!
+//! _Security Considerations_
```
ozgunozerk, line 43: this is not a blocker, but we have the convention `# Security Warning` in our docs. I think following that instead of `_Security Considerations_` should be better. It is still ok if you want to proceed with this one. Approving the PR

PR #334, opened 2025-07-09
Feat/contracttraits

willemneal:
Fixes #??? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation


ozgunozerk:
@willemneal thanks a lot for this!
We gave it a thorough review, and here are our feedback/opinions on the subject:
- Introducing macros is always a tricky subject, because we are effectively creating a DSL on top of Rust, which increases the entry barrier for new adopters
Considering that, the macros introduced should have have the following properties:
- the boilerplate it saves us should worth the complexity of the new DSL
- if possible, it should be intuitive and the users shouldn't need to look-up the documentation to discover how they should use the macro (1-2 examples should be enough for them to adopt the macro)
- again, if possible, the code it hides should be inspected and debugged easily with `cargo expand` command


For this PR, I don't think the criteria are met for the macros introduced. The reasons are:
- the new DSL is much more complex than what we had in `#[default_impl]`
    - `#[default_impl]` is intuitive, because in Rust, developers are accustomed to not provide the default methods for the trait
    - the new DSL proposed here has its own structure and rules, which needs to be learned.
       ```rust
         #[derive_contract(
              AccessControl,
              FungibleToken(ext = FungibleAllowListExt),
              FungibleAllowList(default = ExampleContract),
              FungibleBurnable(ext = FungibleAllowListExt),
          )]
      ```
- the new macro `#[derive_contract]` is working together with `#[contract]` macro, which is really bad in terms of `expand` and `debug` mechanisms. Because, it is unclear what exactly `#[derive_contract]` is doing, due to the expanded code is intertwined with the `#[contract]` macros output as well, which is HUGE (we are talking about 2k to 4k lines of code here)
- on the contrary, `#[default_impl]` macro works really well with `cargo expand` command, and easy to see what's happening inside
- the boilerplate that `#[derive_contract]` saves us, is not significant in our opinion.
    - for every item in `#[derive_contract]` macro -> correlates to 1 line:
       ```rust
       FungibleAllowList(default = ExampleContract),
       ```
    - we can remove 3 lines:
       ```rust
       #[default_impl]
       #[contractimpl]
       impl AccessControl for ExampleContract {}
       ```
    - means, we are only gaining 2 lines per an item in the `#[derive_contract]` macro


When we weigh in the complexity introduced by the new DSL, the "magic behind curtains" introduced by the macro, harder or impossible to debug the contract related code, we are skeptic on this approach.

As Leigh suggested in the discussion, if we can bring the utility of this macro without introducing a new DSL, that would be really great. However, I still think that is technically impossible due to how macros work in Rust.

Note:
- `type Impl = ` is a great workaround, and although it is yet another DSL to learn, it is tolerable imo

examples/nft-royalties/src/contract.rs
```diff
@@ -4,89 +4,92 @@
...
 
         token_id
     }
 
-    pub fn get_royalty_info(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
-        Base::royalty_info(e, token_id, sale_price)
-    }
+    // pub fn get_royalty_info(e: &Env, token_id: u32, sale_price: i128) ->
+    // (Address, i128) {     <Self::royalty_info(e, token_id, sale_price)
+    // }
```
willemneal, line 60:
```suggestion
```

examples/nft-royalties/src/contract.rs
```diff
@@ -4,89 +4,92 @@
...
+            token_id,
+            &receiver,
+            basis_points,
+            &to,
+        );
 
         token_id
     }
 
-    pub fn get_royalty_info(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
```
willemneal, line 51: This seems unneeded since `royalty_info` is already exposed.

PR #335, opened 2025-07-14
chore: prevent test_snapshots from being checked in; make PRs less noisy

willemneal:


ozgunozerk (approved):
I think this will be helpful. We are not utilizing these much, and they are indeed making PRs more noisy

willemneal:
@ozgunozerk Can this be merged?

willemneal:
@brozorec It should be good to go! Will make my other PR from being too noisy!

willemneal:
@brozorec Another Ping!

willemneal:
I agree it has some use, but when making a big refactor it makes it much harder to review. It might make sense to have a special branch and github action that tests if there is a difference with the latest main branch.

@brozorec Okay to merge?

PR #338, opened 2025-07-19
Reorganize docs

brozorec:
- Fixes inline and antora docs according the new structure
- Fixes some examples where `only_role` must be used instead of `has_role`

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ OpenZeppelin Stellar Soroban Contracts is a collection of contracts for the Stel
 - `packages/`: Source code
   - `access/`: Role-based access controls and ownable
   - `contract-utils/`: Utilities for contracts (pausable, upgradeable, crypography, etc.)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,8 +14,7 @@ OpenZeppelin Stellar Soroban Contracts is a collection of contracts for the Stel
 
 - `packages/`: Source code
   - `access/`: Role-based access controls and ownable
-  - `constants/`: Constants shared across token standards
-  - `contract-utils/`: Utilities for contracts (pausable, upgradeable, etc.)
+  - `contract-utils/`: Utilities for contracts (pausable, upgradeable, crypography, etc.)
   - `test-utils/`: Utilities for testing
```
ozgunozerk, line 18: we should add `macros` here I think

ozgunozerk (approved):
looking good to me! It may be possible that we may have missed some fixes, but that's ok, we can always fix them later when we come across them

PR #342, opened 2025-07-30
T rex identity storage and token binder

brozorec:
This PR introduces the following RWA modules following the ERC-3643 design. 

1. Identity Registry Storage: managing user identities and associated country profiles. 
[Check here](https://github.com/OpenZeppelin/stellar-contracts/blob/b576948e199c933d0593b4bd04e95aa84b86b20a/packages/tokens/src/rwa/identity_storage_registry/storage.rs#L1) how about the adopted approach in managing the country attribute.

2. Token Binder Utility: a reusable trait and storage implementation to manage a list of tokens linked to a contract
- to be used with Compliance, Identity Registry Storage and Issuers and ClaimTopics Registry
- added as a placeholder and will need to get refactored because it's inefficient

3. Example implementing Identity Registry Storage

Those are only baseline implementations to kick start the process and the discussion.


packages/tokens/src/rwa/identity_storage_registry/mod.rs
```diff
@@ -0,0 +1,274 @@
...
+pub use storage::*;
+
+use crate::rwa::utils::token_binder::TokenBinder;
+
+/// The core trait for managing basic identities. It is generic over a
+/// `CountryProfile` type, allowing implementers to define what constitutes a
+/// country profile.
+pub trait IdentityRegistryStorage: TokenBinder {
+    /// The specific type used for country profiles in this implementation.
+    type CountryProfile: FromVal<Env, Val>;
```
ozgunozerk, line 15: good idea to make this an associated type and be generic about it by leveraging `FromVal` 👍

packages/tokens/src/rwa/identity_storage_registry/storage.rs
```diff
@@ -0,0 +1,502 @@
...
+            .inspect(|_| {
+                e.storage().persistent().extend_ttl(
+                    &key,
+                    IDENTITY_TTL_THRESHOLD,
+                    IDENTITY_EXTEND_AMOUNT,
+                )
+            })
+            // Unwrap should be always safe, if counting is done correctly,
+            // adding "else" for consistency
+            .unwrap_or_else(|| panic_with_error!(e, IRSError::CountryProfileNotFound));
```
ozgunozerk, line 338: I'd suggest `expect` in this case and removing the comment. That will clearly signal the intention behind the code

packages/tokens/src/rwa/identity_storage_registry/storage.rs
```diff
@@ -0,0 +1,502 @@
...
+    let current_key =
+        IRSStorageKey::CPEnumerable(CPEnumerableKey { account: account.clone(), index });
+    let profile_to_remove = e
+        .storage()
+        .persistent()
+        .get(&current_key)
+        .unwrap_or_else(|| panic_with_error!(e, IRSError::CountryProfileNotFound));
+
+    let count = get_country_profile_count(e, account);
+    // Can't overflow because `profile_to_remove` would panic if count == 0
```
ozgunozerk, line 461: did you mean `get_country_profile_count()` would panic if `count==0` instead of `profile_to_remove`?

ozgunozerk (approved):
have few minor comments, but great work overall 💯
no major concerns 👍

examples/rwa/src/identity_registry_storage.rs
```diff
@@ -0,0 +1,96 @@
...
+    type CountryProfile = CountryProfile;
+
+    fn add_identity(
+        e: &Env,
+        account: Address,
+        identity: Address,
+        initial_profiles: Vec<CountryProfile>,
+        _operator: Address,
+    ) {
+        irs::add_identity(e, &account, &identity, &initial_profiles);
```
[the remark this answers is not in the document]
brozorec: just realized the other meaning hah

packages/tokens/src/rwa/identity_storage_registry/storage.rs
```diff
@@ -0,0 +1,502 @@
...
+///
+/// # Assumptions
+///
+/// 1. **All Profiles are Equal**: The system treats the initial profile and any
+///    subsequently added profiles the same way. They are all stored together in
+///    an enumerable list.
+/// 2. **Efficient but Simple Indexing**: Profiles are stored by a simple index
+///    (0, 1, 2, ...). When a profile is deleted, the last profile in the list
+///    is moved into its place to keep the list compact and gas-efficient. This
+///    means the index of a profile can change.
```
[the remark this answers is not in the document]
brozorec: Good point, I'm still not 100% sure that index changing is a good idea though 🤔

brozorec:
The country data model was changed to reflect the need to have separate country data for individuals and organizations.

packages/tokens/src/rwa/identity_storage_registry/storage.rs
```diff
@@ -440,57 +415,63 @@ pub fn modify_country_profile(e: &Env, account: &Address, index: u32, profile: &
...
+/// * `e` - The Soroban reference.
 /// * `key` - The key required to retrieve the underlying storage.
 fn get_persistent_entry<T: TryFromVal<Env, Val>>(e: &Env, key: &IRSStorageKey) -> Option<T> {
     e.storage().persistent().get::<_, T>(key).inspect(|_| {
         e.storage().persistent().extend_ttl(key, IDENTITY_TTL_THRESHOLD, IDENTITY_EXTEND_AMOUNT);
     })
 }
+
+/// Helper function to Add multiple country profiles without chaching whether
+/// identity already exists and without event emitting.
```
ozgunozerk: it also doesn't check whether the country profiles already exist for the account from what I understand. Should we also add this detail to the inline docs?

packages/tokens/src/rwa/identity_registry_storage/storage.rs
```diff
@@ -0,0 +1,600 @@
...
+pub enum IRSStorageKey {
+    /// Maps account address to identity address
+    Identity(Address),
+    /// Maps an account to its complete identity profile
+    IdentityProfile(Address),
+}
+
+// ################## QUERY STATE ##################
+
+/// Retrieves the stored identity for a given account.
```
ozgunozerk, line 164: I'm confused on the difference between `Account Address` and `Identity Address`. 

If I missed the relative inline comments somewhere else, it's ok. 

If not, can we explain it here also?

packages/tokens/src/rwa/identity_registry_storage/storage.rs
```diff
@@ -0,0 +1,600 @@
...
+    let profile: IdentityProfile =
+        e.storage().persistent().get(&profile_key).expect("identity profile must be already set");
+    e.storage().persistent().remove(&profile_key);
+
+    for country_data in profile.countries {
+        emit_country_data_event(e, CountryDataEvent::Removed, account, &country_data);
+    }
+}
+
+/// Adds multiple country data entries to an existing identity.
```
ozgunozerk, line 403: It is mentioned above in the library level inline docs, but I still think including the statement "does not check for duplicate countries" here is valuable

ozgunozerk (changes requested):
Requested some changes, all of them are minor, good job!

packages/tokens/src/rwa/identity_registry_storage/storage.rs
```diff
@@ -0,0 +1,600 @@
...
+    if profile.countries.len() == 1 {
+        panic_with_error!(e, IRSError::EmptyCountryList)
+    }
+
+    let country_data_to_remove = profile
+        .countries
+        .get(index)
+        .unwrap_or_else(|| panic_with_error!(e, IRSError::CountryDataNotFound));
+
+    profile.countries.remove(index);
```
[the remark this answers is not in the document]
brozorec: It is more efficient, but is it worth it when there could be only up to 15 elements?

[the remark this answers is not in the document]
ozgunozerk: didn't know it didn't exist on `soroban_sdk::Vec`, in this case let's go with `remove()`. As you said, shouldn't make much difference for at most 15 items. 

I think putting a comment above `remove()` line, and having a TODO for changing it with `swap_remove()` in the future if it will be supported by `soroban_sdk::Vec` is a good idea for the future

ozgunozerk (approved):

Issue #344, opened 2025-07-30
🏗️ [Core Feature]: Implement ClaimTopicsAndIssuers module for RWA


ozgunozerk:
### What is the key feature we're aiming to implement?

We have decided to merge ClaimTopicsRegistry and TrustedIssuers modules together into a single contract.

PR #345, opened 2025-07-30
T rex compliance issuers topics

### LINKED ISSUE #343: 🏗️ [Core Feature]: Implement Compliance module for RWA

We are aiming to implement to compliance module as is, without many changes

ozgunozerk:
Important things here are:
- `compliance` and `ClaimTopicsAndIssuers` are not extensions for RWA, but still relevant to RWA and should be in the same folder
- so, I renamed the `extensions` folder as `modules` to signal these are pluggable modules. 
- there are no trait bound requirement for `RWA` such as `Compliance: RWA`, because these modules are designed to be standalone smart contracts instead of extensions on the same RWA contract (the necessary explanations are put as inline docs). 

Fixes #343 and #344 

#### PR Checklist

- [x] Tests
- [x] Documentation


packages/tokens/src/rwa/claim_topics_and_issuers/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// * `e` - Access to the Soroban environment.
+    /// * `issuer` - The address of the trusted issuer's claim issuer contract.
+    /// * `claim_topic` - The claim topic that has to be checked to know if the
+    ///   issuer is allowed to emit it.
+    fn has_claim_topic(e: &Env, issuer: &Address, claim_topic: u32) -> bool;
+}
+
+// ################## ERRORS ##################
+
+#[contracterror]
```
brozorec, line 159: do we define the errors for each submodule or in top level one?

packages/tokens/src/rwa/claim_topics_and_issuers/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+
+// ################## CONSTANTS ##################
+
+const DAY_IN_LEDGERS: u32 = 17280;
+pub const CLAIMS_EXTEND_AMOUNT: u32 = 30 * DAY_IN_LEDGERS;
+pub const CLAIMS_TTL_THRESHOLD: u32 = CLAIMS_EXTEND_AMOUNT - DAY_IN_LEDGERS;
+pub const ISSUERS_EXTEND_AMOUNT: u32 = 30 * DAY_IN_LEDGERS;
+pub const ISSUERS_TTL_THRESHOLD: u32 = ISSUERS_EXTEND_AMOUNT - DAY_IN_LEDGERS;
+pub const MAX_CLAIM_TOPICS: u32 = 15;
+pub const MAX_ISSUERS: u32 = 50;
```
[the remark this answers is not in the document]
ozgunozerk: They are constrained in the in the ethereum implementation, and I think it make sense to bound them in a smart contract environment. There are many functions (getters and setters) that will deal with all these and their cost will increase linearly w.r.t this number.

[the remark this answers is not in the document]
brozorec: Indeed, the more complex tx you make, the more you pay. Nevertheless, the limit in the Solidity implementation is set because there are many cross-contract calls and very quickly, it runs out of gas. We greatly optimized the structure and the flow and reduced significantly the cross-contract calls and I bet we can be way more permissive. I'm curious to see how much better it is :-) I'll give it a thought about how to benchmark it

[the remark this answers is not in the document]
ozgunozerk: Sure, the numbers can be and should be tweaked. I put the numbers just as a reminder so that we should limit them with constants 👍

packages/tokens/src/rwa/compliance/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    // TODO: add `TokenBinder` bound when it is available
+
+    /// Registers a compliance module for a specific hook type.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `hook` - The type of hook to register the module for.
+    /// * `module` - The address of the compliance module contract.
+    fn add_module_to(e: &Env, hook: HookType, module: Address);
```
brozorec, line 58: Wouldn't it be judicious to have a parameter `hooks` with a vector  of `HookType` instead of single?

packages/tokens/src/rwa/compliance/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+
+// ################## COMPLIANCE MODULE TRAIT ##################
+
+/// Trait for compliance modules that can be registered with the modular
+/// compliance system.
+///
+/// Modules implement this trait to provide specific compliance logic for
+/// different hook types. Each module only needs to implement the methods for
+/// hooks it registers for.
+pub trait ComplianceModule {
```
brozorec, line 216: As we differentiate between diff types of hooks, shouldn't we have a `ComplianceModule` per hook instead of one for all of them?

packages/tokens/src/rwa/compliance/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+/// [`Compliance`] trait is not expected to be an extension to a RWA smart
+/// contract, but it is a separate contract on its own. This design allows it to
+/// be shared across many RWA tokens. Note that, there is no `RWA` bound on the
+/// [`Compliance`] trait:
+///
+/// ```rust, ignore
+/// pub trait Compliance       // ✅
+/// pub trait Compliance: RWA  // ❌
+/// ```
+pub trait Compliance {
```
brozorec, line 48: Each of `transferred()`, `created()` and `destroyed()` can be state changing, so we need to add an additional parameter `token: Address` on which to call `require_auth()` and check whether it was bound to `Compliance`.

packages/tokens/src/rwa/compliance/storage.rs
```diff
@@ -0,0 +1,309 @@
...
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `from` - The address that sent the tokens.
+/// * `to` - The address that received the tokens.
+/// * `amount` - The amount of tokens transferred.
+///
+/// # Cross-Contract Calls
+///
+/// Invokes `on_xfer(from, to, amount)` on each registered module.
```
brozorec, line 198:
```suggestion
/// Invokes `on_transfer(from, to, amount)` on each registered module.
```

packages/tokens/src/rwa/compliance/storage.rs
```diff
@@ -0,0 +1,309 @@
...
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `to` - The address that received the newly created tokens.
+/// * `amount` - The amount of tokens created.
+///
+/// # Cross-Contract Calls
+///
+/// Invokes `on_mint(to, amount)` on each registered module.
```
brozorec, line 225:
```suggestion
/// Invokes `on_created(to, amount)` on each registered module.
```

packages/tokens/src/rwa/compliance/storage.rs
```diff
@@ -0,0 +1,309 @@
...
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `from` - The address from which tokens were destroyed.
+/// * `amount` - The amount of tokens destroyed.
+///
+/// # Cross-Contract Calls
+///
+/// Invokes `on_burn(from, amount)` on each registered module.
```
brozorec, line 252:
```suggestion
/// Invokes `on_destroyed(from, amount)` on each registered module.
```

packages/tokens/src/rwa/compliance/storage.rs
```diff
@@ -0,0 +1,309 @@
...
+///
+/// # Cross-Contract Calls
+///
+/// Invokes `on_xfer(from, to, amount)` on each registered module.
+pub fn transferred(e: &Env, from: Address, to: Address, amount: i128) {
+    let modules = get_modules_for_hook(e, HookType::Transfer);
+
+    // Call each registered module
+    for module_address in modules.iter() {
+        let _result: () = e.invoke_contract(
```
brozorec, line 204: Why not using a client call instead of the low-level `invoke_contract`? `invoke_contract` also works here, but it seems more suitable for cases where the arguments are unknown at compile time, while here that's not the case.

packages/tokens/src/rwa/compliance/mod.rs
brozorec, line 216: Each module is supposed to be bound to a specific `Compliance` and the its methods called only the bound `Compliance`. Should we add methods for that to the trait and an additional argument `compliance: Address` to the state-modifying methods?

brozorec (changes requested):

packages/tokens/src/rwa/claim_topics_and_issuers/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+///
+/// [`ClaimTopicsAndIssuers`] trait is not expected to be an extension to a RWA
+/// smart contract, but it is a separate contract on its own. This design allows
+/// it to be shared across many RWA tokens. Note that, there is no `RWA` bound
+/// on the [`ClaimTopicsAndIssuers`] trait:
+///
+/// ```rust, ignore
+/// pub trait ClaimTopicsAndIssuers       // ✅
+/// pub trait ClaimTopicsAndIssuers: RWA  // ❌
+/// ```
```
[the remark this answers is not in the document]
ozgunozerk: or maybe we can explain the details in the openzeppelin docs?

packages/tokens/src/rwa/compliance/mod.rs
ozgunozerk: I thought about it as well. There are 2 reasons I decided against it:
- what if the module is already added to one of the given hooks, but not the others? Should we give an error? Or just add the module to other hooks and silently ignore the hook that the module is already registered to
- I reckon adding and removing modules to/from hooks will not be an everyday operation like transfers. Additionally, most of the modules that will be added to hooks, will be specifically crafted towards a single hook type. So, complicating the logic and the user experience for the sake of this optimization didn't sound so great to me (we don't gain much from this optimization, they can call this function multiple times, and the workflow will be simpler).

packages/tokens/src/rwa/compliance/mod.rs
ozgunozerk: let's continue this discussion from Slack as well

packages/tokens/src/rwa/compliance/storage.rs
ozgunozerk: good call, thanks!

packages/tokens/src/rwa/compliance/mod.rs
ozgunozerk: let's also discuss this on Slack

ozgunozerk: agreed, let's keep this convo open and resolve it once token_binder PR is merged

packages/tokens/src/rwa/claim_topics_and_issuers/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+pub trait ClaimTopicsAndIssuers {
```
brozorec, line 19: As per the proposed design, there should be a fn that returns all claim topics and their trusted issuers so that we get all the data we need in `is_verified()` in one go, instead of looping through all claim topics and making N another calls to fetch the corresponding trusted issuers.

PR #346, opened 2025-08-01
Implement initial vault standard functionality

### LINKED ISSUE #148: 🏗️ [Core Feature]: ERC4626 Tokenized Vault Standard Implementation for Soroban

Hello,

Following an initial email discussion and confirmation, we are submitting this core feature request to propose to collaborate on an open-source implementation of the ERC4626 Tokenized Vault Standard for Soroban/Stellar.

We (Team Sentinel) propose working closely with OpenZeppelin to develop a standardized implementation of ERC4626 for Soroban, based on the following:

1. Code Refinement and Standardization:

- Refactor and improve our current Soroban Vault implementation to meet OpenZeppelin’s standards.
https://github.com/SentinelFi/soroban_vault

2. Alignment with Official ERC4626:

- Adhere closely to OpenZeppelin’s existing ERC4626 implementation for EVM.
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol

3. Integration with OpenZeppelin’s Soroban Standards:

- Build atop OpenZeppelin’s developed Fungible Token implementation for Soroban.
https://github.com/OpenZeppelin/stellar-contracts/tree/main/packages/tokens/fungible

4. Open-Source and Developer Tooling:

- Fully open-source the implementation and integrate it into OpenZeppelin’s repositories and the Stellar Contracts Wizard.
- Include comprehensive documentation, comments for security best practices, tests and usage guidelines.
https://wizard.openzeppelin.com/stellar

**Benefits to the Stellar Ecosystem:**
This would provide Soroban developers with a standardized, secure foundation for building token vaults, lending protocols, and other DeFi applications that require tokenized asset management.

We invite all interested parties to discuss:

- Technical considerations (e.g., deviations from EVM due to Soroban’s design).
- Collaboration processes (e.g., PR workflow, testing).
- Additional features or safeguards.

Let’s finalize the approach in this thread before proceeding with a pull request.

jsmaxi:
[Draft] Fixes #148  

**ERC4626 Vault Standard Implementation - Review Request**

Hello,

We are seeking initial feedback on our draft ERC4626 vault standard implementation. While this is not the final PR version, we would appreciate your input on our current approach and any recommendations you might have.

Specific questions:

1. **Vault Extension Placement**: Have we positioned the vault extension in the appropriate location within the codebase?

2. **Math Library**: What are your thoughts on using this Soroban fixed point math library (https://github.com/script3/soroban-fixed-point-math/) for muldiv operations?

3. **Utility Functions**: Where would you recommend placing math utility functions to ensure they are accessible to token contracts in future implementations? (Currently located in _math.rs_)

Remaining tasks:

- Implement vault examples with constructor and admin/ownable patterns.
- Add comprehensive documentation.
- Include detailed code comments.
- Expand test coverage.
- Code optimization and refinements.

Our standard closely follows OpenZeppelin's EVM implementation and OpenZeppelin's Stellar contract architecture, referencing:
- [EIP-4626 Specification](https://eips.ethereum.org/EIPS/eip-4626)
- [OpenZeppelin ERC4626 Implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol)
- [IERC4626 Interface](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC4626.sol)
- [OpenZeppelin Math Utilities](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol)

We welcome any comments, suggestions, or reviews from your team.

#### PR Checklist

- [ ] Tests
- [x] Documentation


brozorec:
Hey @jsmaxi that looks very good 👍 let us review it in depth

packages/tokens/src/fungible/extensions/vault/storage.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub fn preview_mint(e: &Env, shares: i128) -> i128 {
+        _convert_to_assets(e, shares, Rounding::Ceil)
+    }
+
+    pub fn max_withdraw(e: &Env, owner: Address) -> i128 {
+        _convert_to_assets(e, Base::balance(e, &owner), Rounding::Floor)
+    }
+
+    pub fn preview_withdraw(e: &Env, assets: i128) -> i128 {
+        _convert_to_shares(e, assets, Rounding::Ceil)
```
ozgunozerk, line 70: shouldn't this be `_convert_to_assets` instead?

packages/tokens/src/fungible/extensions/vault/storage.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub fn total_assets(e: &Env) -> i128 {
+        let token_client = token::Client::new(&e, &Vault::query_asset(&e));
+        token_client.balance(&e.current_contract_address())
+    }
+
+    pub fn convert_to_shares(e: &Env, assets: i128) -> i128 {
+        _convert_to_shares(e, assets, Rounding::Floor)
+    }
+
+    pub fn convert_to_assets(e: &Env, shares: i128) -> i128 {
```
ozgunozerk, line 45: `convert_to_assets`, `preview_mint`, and `preview_redeem` share the exact same implementation. 

I think there is no need to expose the same functionality under 3 different names. Will their implementation diverge at some point?

ozgunozerk: the same goes for `_convert_to_shares` function body that is shared by multiple functions

packages/tokens/src/fungible/extensions/vault/mod.rs
```diff
@@ -0,0 +1,79 @@
...
+use soroban_sdk::{symbol_short, Address, Env};
+
+pub use storage::Vault;
+
+use crate::fungible::FungibleToken;
+
+/// Vault Trait for Fungible Token
+/// TODO: describe trait, functions, arguments
+
+pub trait FungibleVault: FungibleToken<ContractType = Vault> {
```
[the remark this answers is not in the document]
jsmaxi: ERC4626 inherits from ERC20 (balance and total supply of share tokens)
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol 
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/IERC20.sol

packages/tokens/src/fungible/extensions/vault/storage.rs
jsmaxi: Based on EVM standard implementation:    

```
function previewDeposit(uint256 assets) public view virtual returns (uint256) {
        return _convertToShares(assets, Math.Rounding.Floor);
}
```

packages/tokens/src/fungible/extensions/vault/storage.rs
jsmaxi: It is implementing a standard interface with minimal code duplication. For example, ```preview_mint``` and ```preview_redeem``` require different ```Rounding``` parameters. Using a macro or other solution for that would not improve code readability, imho.

packages/tokens/src/fungible/extensions/vault/storage.rs
```diff
@@ -0,0 +1,242 @@
...
+            .checked_add(_decimals_offset())
+            .unwrap_or_else(|| panic_with_error!(e, FungibleTokenError::MathOverflow))
+    }
+}
+
+/**
+ * Internal conversion function (from assets to shares) with support for rounding direction.
+ * assets.mulDiv(totalSupply() + 10 ** _decimalsOffset(), totalAssets() + 1, rounding)
+ */
+fn _convert_to_shares(e: &Env, assets: i128, rounding: Rounding) -> i128 {
```
brozorec, line 163: The internal functions can be included in the Vault impl as well. I see the idea to follow the Sol convention of prefixing them with "_", but here we might prefer suffixing with `_no_auth` for the changing state ones and `convert_to_shares_with_rounding` (or smth shorter).

brozorec (changes requested):
Thanks for your contribution @jsmaxi, great job 🙌 I'm happy with moving on 👍

Cargo.toml
```diff
@@ -45,6 +45,7 @@ syn = { version = "2.0", features = ["full"] }
 soroban-test-helpers = "0.2.3"
 hex-literal = "0.4.1"
 ed25519-dalek = "2.1.1"
+soroban-fixed-point-math = "1.3.0"
```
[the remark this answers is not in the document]
jsmaxi: The person who actually suggested that I use that library is Silence (https://github.com/silence48), he is a very experienced Stellar/Soroban developer. I was told by him that it was audited, but I suppose I also need to find official audit document to back up those claims.

packages/tokens/src/fungible/extensions/vault/storage.rs
jsmaxi: Will refactor, thank you.

packages/tokens/src/fungible/math.rs
```diff
@@ -0,0 +1,23 @@
...
+ * Relies on https://github.com/script3/soroban-fixed-point-math/ math library.
+ */
+pub fn muldiv(e: &Env, x: i128, y: i128, denominator: i128, rounding: Rounding) -> i128 {
+    match rounding {
+        Rounding::Floor => x.fixed_mul_floor(e, &y, &denominator),
+        Rounding::Ceil => x.fixed_mul_ceil(e, &y, &denominator),
+    }
+}
+
+// !! TODO: decide where to place math utils for re-use
```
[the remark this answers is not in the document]
ozgunozerk: `contract-utils` sounds like a good place to me as well 👍

packages/tokens/src/fungible/extensions/vault/storage.rs
ozgunozerk: Ok let's eliminate the ones with `Rounding` parameters, didn't notice that diff while taking a quick glance. 

Let me ask the question in a more polished way:

both `convert_to_shares()` and `preview_deposit()` share the exact same body, which is:
` _convert_to_shares(e, assets, Rounding::Floor)`

It sounds counter-intuitive to me to have 2 different functions that have the exact same body. 

> It is implementing a standard interface with minimal code duplication.

My concern is not code duplication here, the shared function body is already made re-usable, no issues there.

> Using a macro or other solution for that would not improve code readability, imho.

Again, it is not about code readability. I don't think macros will help either in this case. 

---

My question is more on the high-level. 

> Will their implementation diverge at some point?

If so, it is perfectly ok to keep as is. But if they are meant to do the same job, having 2 different functions does not make sense from technical point of view, however it may make sense in terms of UX.

One follow-up question from my end would be, do we need to expose `convert_to_shares()`? Is this function expected to be public and use widely? Would `preview_deposit()` be enough on its own to provide a good UX?

The same question applies to `preview_redeem()` and `convert_to_assets()`.

packages/tokens/src/fungible/extensions/vault/storage.rs
ozgunozerk: I was confused due to these:

```rust
    pub fn max_withdraw(e: &Env, owner: Address) -> i128 {
        _convert_to_assets(e, Base::balance(e, &owner), Rounding::Floor)
    }

    pub fn preview_withdraw(e: &Env, assets: i128) -> i128 {
        _convert_to_shares(e, assets, Rounding::Ceil)
    }
```

At my first quick glance, I thought both `withdraw` related functions should call the same variant. Got it now, thanks!

Cargo.toml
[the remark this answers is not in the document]
ozgunozerk: In this case, I support the approach of:

> it is possible strip away the unnecessary math library code and use only the required logic internally. This would eliminate potential concerns about library compatibility with future updates, security, other dependencies, etc.

examples/fungible-vault/src/contract.rs
```diff
@@ -0,0 +1,144 @@
...
+#[contractimpl]
+impl FungibleVault for ExampleContract {
+    // Allows override of public vault functions.
+
+    // While overriding is technically possible, most implementations should stick to the
+    // pattern of setting the asset address in the constructor and returning it from
+    // instance storage, as this maintains predictability and composability that
+    // other contracts and users expect from vaults.
+    // IMPORTANT: If overriding query_asset, you MUST also override all other functions that depend on it.
+    // Failure to override these functions will result in inconsistent behavior.
```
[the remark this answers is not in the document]
ozgunozerk: I've given some thought on this. 

Here are some points:
- consistency across the library is quite important for us. This does not mean we won't change any prior decisions. This means, if we choose to go with the approach suggested here, we should also do a considerable refactor to all other parts of the library as well to utilize generics in such a way. Because, calling internal storage functions from other storage functions is not unique to `vault` extension (although, I agree `vault` extension utilizes this more compared to other parts of the library). 
- Our current approach to flexibility and ease of use can be summarized like this: we provide sensible defaults, so for the general use case, one does not have to mess with the code much. For every other scenario that doesn't reside within the general use case boundaries, the correct implementation is the responsibility of the developer.
- Coming back to the `query_asset` function discussion, the necessity to override it, seemed to me like a very very low probability. And just to cover that minuscule case, further complicating the code for the general use-case doesn't sound like a great choice to me. The problem of other functions getting affected is also applicable to other tokens and extensions as well (for example, `total_supply` function is also used internally in the `fungible` token's storage functions, yet we don't warn the user on if they override the `total_supply` function, they should also override other functions that are using it. Because, overriding `total_supply` is not expected, and if someone has that specific needs, they probably already know what they are doing 
- Re-iterating on our current approach: we provide `storage.rs` functions as building blocks (both high and low-level). Whether to use these building blocks or not is up to the developer. So, I think we should't introduce generics, and we should also get rid of the `FungibleVaultInternal` to keep consistency. The developers have all the tools they need to override the trait by utilizing storage.rs functions.
- If we think `FungibleVaultInternal` is a good idea (separation of lower level functions for customization, and high-level external functions automatically changing their behavior accordingly), then it is also applicable to at least some other parts of our library. This means, a huge refactor, and an architectural design change for our library. I'm not against it by any means, and definitely open for discussing it. However, it is out of the scope of this PR. If you have strong opinions on this subject, let's open another issue/discussion and continue from there. We can even set a sync meeting (would love to meet you).

TL;DR:
I think we should get rid of the generics and the `FungibleVaultInternal` trait, and follow the same approach as other extensions/tokens from our library. I acknowledge that `vault` extension is utilizing the connection between internal functions a bit more, but I don't see this as an impediment towards not following the same design across our library.

packages/tokens/src/fungible/extensions/vault/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn max_mint(e: &Env, receiver: Address) -> i128;
+    fn preview_mint(e: &Env, shares: i128) -> i128;
+    fn mint(e: &Env, shares: i128, caller: Address, receiver: Address) -> i128;
+    fn max_withdraw(e: &Env, owner: Address) -> i128;
+    fn preview_withdraw(e: &Env, assets: i128) -> i128;
+    fn withdraw(e: &Env, assets: i128, caller: Address, receiver: Address, owner: Address) -> i128;
+    fn max_redeem(e: &Env, owner: Address) -> i128;
+    fn preview_redeem(e: &Env, shares: i128) -> i128;
+    fn redeem(e: &Env, shares: i128, caller: Address, receiver: Address, owner: Address) -> i128;
+}
```
brozorec, line 35: Rename `caller: Address` to `operator: Address` and move it as a last fn parameter for a consistency with other interfaces

packages/tokens/src/fungible/extensions/vault/storage.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    // ################## CHANGE STATE ##################
+
+    pub fn deposit<T: FungibleVault>(
+        e: &Env,
+        assets: i128,
+        caller: Address,
+        receiver: Address,
+    ) -> i128 {
+        caller.require_auth();
```
brozorec, line 91: Remove `require_auth()` here and in all the functions below. The authorization should be left for the implementer. 

The implementer can use `ownable` or `access_control` that provide functions to handle authorizations (or any other custom approach). Requiring auth on `caller` here has no securing effect, because anyone can still call this function if there're no additional check s on `caller`'s role.

examples/fungible-vault/src/contract.rs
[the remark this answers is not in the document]
brozorec: I agree with @ozgunozerk that we should keep it straight and consistent and remove the internal trait and generics. We can place the those low-level functions directly in the Vault impl.

brozorec (changes requested):
@jsmaxi there are a couple of relatively straightforward changes and imo we can aim for including your work in the scope for an audit that starts next week. Do you think it's possible we can finalize this PR by then? Ofc we can jump in and help with tests, docstrings or any other thing. Plz let us know

packages/tokens/src/fungible/extensions/vault/mod.rs
jsmaxi: Renamed and moved

packages/tokens/src/fungible/extensions/vault/storage.rs
jsmaxi: Removed

examples/fungible-vault/src/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+    type ContractType = Vault;
+
+    // Allows override of decimals and other base functions.
+
+    fn decimals(e: &Env) -> u32 {
+        Vault::decimals(e)
+    }
+}
+
+#[contractimpl]
```
ozgunozerk, line 42: `#[default_impl]` support for this would be good. I can push that change as well, this is just a reminder

ozgunozerk (approved):
LGTM!

I think it is time we can address the documentation and remaining small TODO's. 

What do you think @brozorec ?

@jsmaxi if you want to tackle the inline documentation and remaining TODOs, feel free to do so. If you have other things on your plate, I can gladly complete the remaining tasks. Let me know your preference :)

examples/fungible-vault/src/contract.rs
ozgunozerk: Sounds good! 

> it's great to collaborate with you both!

Likewise 🥂 

Thanks a lot again for all this 🚀

examples/fungible-vault/src/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    fn max_redeem(e: &Env, owner: Address) -> i128 {
+        Vault::max_redeem(e, owner)
+    }
+
+    fn preview_redeem(e: &Env, shares: i128) -> i128 {
+        Vault::preview_redeem(e, shares)
+    }
+
+    fn deposit(e: &Env, assets: i128, caller: Address, receiver: Address) -> i128 {
```
jsmaxi, line 94: Ah in the example contract, after the recent change, the `withdraw` `deposit` `redeem` and `mint` functions must now pass the `operator` address as the last parameter instead of `caller` (update parameter order and naming).

brozorec (approved):

PR #350, opened 2025-08-07
feat: use proposed `contracttrait` type

willemneal:
This also simplifies Upgrade to depend on Ownable.

A continuation of #334

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -406,6 +406,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e7c1832837b905bbfb5101e07cc24c8deddf52f93225eee6ead5f4d63d53ddcb"
 dependencies = [
  "const-oid",
+ "pem-rfc7468",
  "zeroize",
 ]
 
@@ -476,6 +477,7 @@ dependencies = [
  "elliptic-curve",
  "rfc6979",
  "signature",
+ "spki",
 ]
 
 [[package]]
@@ -521,6 +523,8 @@ dependencies = [
  "ff",
  "generic-array",
  "group",
+ "pem-rfc7468",
+ "pkcs8",
  "rand_core 0.6.4",
  "sec1",
  "subtle",
@@ -841,7 +845,9 @@ dependencies = [
  "cfg-if",
  "ecdsa",
  "elliptic-curve",
+ "once_cell",
  "sha2",
+ "signature",
 ]
 
 [[package]]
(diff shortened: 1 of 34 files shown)


ozgunozerk:
I'll post some of the discussion points below that happened through DM. So, that @brozorec and others can chime in.

examples/fungible-allowlist/src/contract.rs
```diff
@@ -6,66 +6,66 @@
...
-        access_control::grant_role_no_auth(e, &admin, &manager, &symbol_short!("manager"));
+        Self::grant_role_no_auth(e, admin, manager, &symbol_short!("manager"));
 
         // Allow the admin to transfer tokens
-        AllowList::allow_user(e, &admin);
+        Self::allow_user_no_auth(e, admin);
 
         // Mint initial supply to the admin
-        Base::mint(e, &admin, initial_supply);
+        Self::internal_mint(e, admin, initial_supply);
```
ozgunozerk, line 39: Me:
> `internal_mint` is a new addition?

Willem:
> `mint` was public but added an `internal_mint` to the `FT` trait with `#[internal]`. This way if you you wanted to swap of the default implementation it would be guaranteed to have it.

examples/nft-royalties/src/test.rs
```diff
@@ -16,15 +16,14 @@ fn test_default_royalty() {
...
-
     e.mock_all_auths();
+    let client = create_client(&e, &admin, &manager);
 
     // Mint a token
     let token_id = client.mint(&admin);
 
     // Check royalty info (should use default 10%)
-    let (receiver, amount) = client.get_royalty_info(&token_id, &1000);
+    let (receiver, amount) = client.royalty_info(&token_id, &1000);
```
ozgunozerk, line 26: Me:
> `get_royalty_info` is renamed to `royalty_info`, debatable…

Willem:
> It wasn’t renamed because `royalty_info` was part of the trait and thus exposed already. So there were two functions that did the same thing. So I’m open to renaming.

examples/fungible-allowlist/src/contract.rs
```diff
@@ -6,66 +6,66 @@
...
     }
 }
 
-#[default_impl]
 #[contractimpl]
 impl FungibleToken for ExampleContract {
-    type ContractType = AllowList;
+    type Impl = AllowList;
```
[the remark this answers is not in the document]
ozgunozerk: I still think `ContractType` is more self-explanatory than an abstract name such as `Impl`. I'd prefer to keep it as `ContractType`

examples/fungible-allowlist/src/contract.rs
ozgunozerk: I'm not sure I fully get it, could you elaborate on that a bit more @willemneal ?

examples/fungible-allowlist/src/contract.rs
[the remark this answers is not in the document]
willemneal: How about `ContractImpl`? `Type` is also broad and it already is a type. What it is is the implementation of the contract so `ContractImpl` seems appropriate to me.

PR #355, opened 2025-08-27
Smart account

brozorec:
Fixes #??? 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/accounts/src/lib.rs
+++ b/packages/accounts/src/lib.rs
@@ -1,0 +1,3 @@
+pub mod policies;
+pub mod smart_account;
+pub mod verifiers;
--- a/packages/accounts/src/smart_account/storage.rs
+++ b/packages/accounts/src/smart_account/storage.rs
@@ -14,11 +14,13 @@
 //!
 //! ### **What** - Context Rules  
 //! - Rules define authorization requirements for specific contexts (contract
-//!   calls, deployments)
+//!   calls, deployments).
+//! - Each rule must contain at least one signer or one policy and can have an
+//!   optional expiration (`valid_until`) defined by a ledger sequence.
 //! - Multiple rules can exist for the same context type with different signer
-//!   sets and thresholds
+//!   sets and policies.
 //! - Context types: `Default` (any context), `CallContract(Address)`,
-//!   `CreateContract(BytesN<32>)`
+//!   `CreateContract(BytesN<32>)`.
 //! - Each rule specifies required signers and optional policies
 //!
 //! ### **How** - Policies
@@ -46,16 +48,17 @@
 //!
 //! When verifying a context with provided signers:
 //!
-//! I. Get all rules for the specific context type, plus default rules
-//! II. For each rule (start from the last stored):
-//!     1. Identify authenticated signers out of all stored signers
-//!     2.a. If there are policies, verify they can be enforced
-//!         - If all policies can be satisfied, return success
-//!         - Otherwise, move to the next rule
+//! I. Get all non-expired rules for the specific context type, plus default
+//! rules. II. For each rule (iteration starts from the last-stored, which
+//! prevails in    case of conflicting non-expired rules):
+//!     1. Identify authenticated signers out of all stored signers.
+//!     2.a. If there are policies, verify they can be enforced:
+//!         - If all policies can be satisfied, return success.
+//!         - Otherwise, move to the next rule.
 //!     2.b. If there are no policies:
-//!         - Return success only if all signers are authenticated
-//!         - Otherwise, move to the next rule
-//! III. If no rule satisfies above conditions, authorization fails
+//!         - Return success only if all signers are authenticated.
+//!         - Otherwise, move to the next rule.
+//! III. If no rule satisfies the above conditions, authorization fails.
 //!
 //! ## Benefits
 //!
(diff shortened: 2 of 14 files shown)


packages/accounts/src/smart_account/storage.rs
```diff
@@ -0,0 +1,368 @@
...
+//! II. For each rule (iteration starts from the last-stored, which prevails in
+//! case of conflicting non-expired rules):
+//!     1. Identify authenticated signers out of all stored signers.
+//!     2.a. If there are policies, verify they can be enforced:
+//!         - If all policies can be satisfied, return success.
+//!         - Otherwise, move to the next rule.
+//!     2.b. If there are no policies:
+//!         - Return success only if all signers are authenticated.
+//!         - Otherwise, move to the next rule.
+//! III. If no rule satisfies the above conditions, authorization fails.
```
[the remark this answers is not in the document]
ozgunozerk: i meant, `if any role does not satisfy the above conditions, authorization fails` -> is that what you mean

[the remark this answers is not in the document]
brozorec: talking about context rules, not roles

packages/accounts/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+//! policies:
+//!
+//! - Set authorization requirements for specific context types (contract calls,
+//!   deployments)
+//! - Must contain at least one signer or one policy
+//! - Support optional expiration via ledger sequence (`valid_until`)
+//! - Allow multiple rules for the same context with different configurations
+//! - Evaluated in reverse chronological order (most recent first)
+//!
+//! Supported context types:
```
ozgunozerk, line 34: I'm still confused about these, 1-2 examples for especially `Default` and `CallContract` would be useful on when to use which. No need for elaborate examples, but smt like (i.e. blah blah)

packages/accounts/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+//!
+//! The matching algorithm for authorization:
+//!
+//! 1. Retrieve all non-expired rules for the context type, plus defaults
+//! 2. For each rule (newest first):
+//!    - Filter authenticated signers from rule's signer set
+//!    - If policies exist: verify all are enforceable
+//!    - If no policies: require all signers authenticated
+//!    - First matching rule succeeds
+//! 3. If no rule matches, authorization fails
```
ozgunozerk, line 80: so if only 1 rule matches, authorization succeeds? I don't think so?

Maybe something like: "In order for authorization to pass, all rules must pass"

packages/accounts/src/policies/simple_threshold.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    if params.threshold == 0 || params.threshold > context_rule.signers.len() {
+        panic_with_error!(e, SimpleThresholdError::InvalidThreshold)
+    }
+
+    e.storage().persistent().set(
+        &SimpleThresholdStorageKey::AccountContext(smart_account.clone(), context_rule.id),
+        &params.threshold,
+    );
+}
```
ozgunozerk, line 203: identical function bodies with `set_threshold` (except `params.threshold` vs `threshold`).

Should be possible to re-use the most of the code

packages/accounts/src/lib.rs
brozorec: no, on the contrary: on the 1st rule that's satisfied, we consider the context validated

packages/accounts/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+//!     signers: [delegated_signer_1, delegated_signer_2],
+//!     policies: None,
+//! }
+//!
+//! // Regular operations: 3-of-5 threshold with spending limits
+//! Context Rule {
+//!     name: "Regular",
+//!     context_type: CallContract(token_address),
+//!     signers: [signer_1, signer_2, signer_3, signer_4, signer_5],
+//!     policies: [threshold_policy, spending_limit_policy],
```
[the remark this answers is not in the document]
brozorec: added a readme, lmk if it clarifies the picture

packages/accounts/src/smart_account/storage.rs
```diff
@@ -0,0 +1,999 @@
...
+/// * `context_rule_type` - The type of context rules to retrieve.
+pub fn get_context_rules(e: &Env, context_rule_type: &ContextRuleType) -> Vec<ContextRule> {
+    let ids_key = SmartAccountStorageKey::Ids(context_rule_type.clone());
+    let ids: Vec<u32> = get_persistent_entry(e, &ids_key).unwrap_or(Vec::new(e));
+
+    let mut rules = Vec::new(e);
+    for id in ids.iter() {
+        if e.storage().persistent().has(&SmartAccountStorageKey::Meta(id)) {
+            rules.push_back(get_context_rule(e, id));
+        }
```
ozgunozerk, line 264: It is quite inefficient to query the database twice with the same key:
- one for `has()`
- one for `get_context_rule()`

One option may be to write a low-level helper function, that is same with `get_context_rule`, but returns `Option` instead. 

This way:
- `get_context_rule` can utilize this, and if the return value is `None`, it can still panic.
- `get_context_rules` can also utilize this, and simply can use `unwrap_or()` like logic instead of querying the storage 2 times

packages/accounts/src/lib.rs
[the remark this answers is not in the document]
ozgunozerk: somewhat, but still unclear imo. 

Another step before `add_policy`, such as this may be clear enough:

```

Step: Create Policy (optional)
- create a new contract using `Policy` trait, and implement its methods
- _example code here_


Step: Add The Policy 
- add the policy by calling the function `add_policy`, to which we will give the Address of the policy smart contract

```

This will, I guess, make it crystal clear on how one should interact with policies

ozgunozerk (approved):

PR #383, opened 2025-09-29
recover balance

### LINKED ISSUE #377: Missing Same-identity Check in recovery_address Enables Malicious Recovery
We should to the identity verification for the both new and the lost wallet. 

Since we will retrieve the identities for these wallets, we probably won't need the `investor_onchain_id` as a parameter anymore.

- After a successful recovery, consider updating Identity Registry Storage to sever the association for lost_wallet. That means:
- Remove the mapping (remove_identity(lost_wallet, operator)) or
- Keep the mapping but mark the account frozen and/or flagged as “recovered” in a dedicated registry bit.

ozgunozerk:
Fixes #377 

#### PR Checklist

- [x] Tests
- [x] Documentation

Note: builds on top of `frozen-status-fix` branch, should be merged after #380, hence draft

### DIFF
--- a/examples/rwa/src/test.rs
+++ b/examples/rwa/src/test.rs
@@ -31,7 +31,7 @@ fn bind_max() {
     assert_eq!(client.linked_tokens().len(), 200)
 }
 
-// TODO: add test for checking `recovery_address` fails when contract is paused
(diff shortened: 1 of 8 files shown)


ozgunozerk:
closed in favor of #432

PR #424, opened 2025-10-02
recover identity in IRS

brozorec:
This PR add `recover_identity` in `identity_registry_storage`

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/src/rwa/identity_registry_storage/storage.rs
+++ b/packages/tokens/src/rwa/identity_registry_storage/storage.rs
@@ -1,4 +1,4 @@
-/// # How Country Data Work
+/// ## How Country Data Work
 ///
 /// Instead of a simple, single country code, this system treats an account's
 /// jurisdictional ties as a collection of "Country Data." Each country data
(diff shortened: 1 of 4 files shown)


packages/tokens/src/rwa/identity_registry_storage/storage.rs
```diff
@@ -435,6 +435,77 @@ pub fn remove_identity(e: &Env, account: &Address) {
...
+
+    // Recover identity profile
+    let old_profile_key = IRSStorageKey::IdentityProfile(old_account.clone());
+    let new_profile_key = IRSStorageKey::IdentityProfile(new_account.clone());
+
+    let profile: IdentityProfile = e
+        .storage()
+        .persistent()
+        .get(&old_profile_key)
+        .expect("identity profile must be already set");
```
ozgunozerk, line 501: 2 concerns from my end:
- `old_account` is provided as an argument, which means, there can be user error of providing the wrong account. This means, instead of `expect()`, I think it may be better to panic with a specific error instead
- if for somehow, this part panics, I think the above storage modifications won't be rollbacked. I have to double-check on this (whether the whole function reverts/rollbacks if a panic happens). But if what I remember is correct (there will be storage side-effects after the panic), then we should first do the necessary checks above, then modify the storage only after we are sure nothing will panic

brozorec:
added a new function in the IRS trait to query recovery target that should be used in #383

ozgunozerk (approved):

PR #426, opened 2025-10-03
compliance module rework

### LINKED ISSUE #425: 🏗️ [Core Feature]: Update Compliance Implementation to Support Multi-Token Architecture

Compliance trait methods (transferred, created, destroyed, can_transfer, can_create) need to accept a contract: Address parameter identifying the calling RWA token. This enables a single compliance contract to serve multiple RWA tokens by allowing modules to maintain token-specific state and business logic.

Update the implementation in packages/tokens/src/rwa/compliance/lib.rs to accept and forward this parameter to registered modules when invoking hooks.

ozgunozerk:
Fixes #425 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/src/rwa/compliance/mod.rs
+++ b/packages/tokens/src/rwa/compliance/mod.rs
@@ -53,6 +53,14 @@ pub enum ComplianceHook {
 /// pub trait Compliance       // ✅
 /// pub trait Compliance: RWA  // ❌
 /// ```
+///
+/// # Multi-Token Support
+///
+/// To enable a single compliance contract to serve multiple RWA tokens, all
+/// hook functions accept a `contract` parameter identifying the calling RWA
+/// token. This allows compliance modules to maintain separate state and apply
+/// different business logic per token (e.g., token-specific transfer limits,
+/// per-token balance tracking).
 #[contractclient(name = "ComplianceClient")]
 pub trait Compliance {
     /// Registers a compliance module for a specific hook type.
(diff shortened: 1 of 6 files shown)


packages/tokens/src/rwa/compliance/storage.rs
```diff
@@ -201,16 +201,17 @@ pub fn remove_module_from(e: &Env, hook: ComplianceHook, module: Address) {
...
-pub fn transferred(e: &Env, from: Address, to: Address, amount: i128) {
+pub fn transferred(e: &Env, from: Address, to: Address, amount: i128, contract: Address) {
     let modules = get_modules_for_hook(e, ComplianceHook::Transferred);
 
     for module_address in modules.iter() {
         let client = ComplianceModuleClient::new(e, &module_address);
-        client.on_transfer(&from, &to, &amount);
+        client.on_transfer(&from, &to, &amount, &contract);
     }
 }
```
brozorec, line 216:
```suggestion
pub fn transferred(e: &Env, from: Address, to: Address, amount: i128, token: Address) {
  token.require_auth();
  // also check token is bound
  let modules = get_modules_for_hook(e, ComplianceHook::Transferred);

  for module_address in modules.iter() {
    let client = ComplianceModuleClient::new(e, &module_address);
        client.on_transfer(&from, &to, &amount, &token);
  }
}
```

I think we have to enforce authorization here, because this should be callable only by bound tokens.
I also suggest we rename `contract` to `token`, because this is not just some kind of a contract, but a token specifically.

Same comment goes for all the other functions below.

packages/tokens/src/rwa/compliance/mod.rs
```diff
@@ -53,6 +53,14 @@ pub enum ComplianceHook {
...
+///
+/// # Multi-Token Support
+///
+/// To enable a single compliance contract to serve multiple RWA tokens, all
+/// hook functions accept a `contract` parameter identifying the calling RWA
+/// token. This allows compliance modules to maintain separate state and apply
+/// different business logic per token (e.g., token-specific transfer limits,
+/// per-token balance tracking).
 #[contractclient(name = "ComplianceClient")]
 pub trait Compliance {
```
brozorec, line 65: Compliance should be a `TokenBinder`, right?

brozorec (changes requested):

packages/tokens/src/rwa/compliance/storage.rs
ozgunozerk: makes sense, will address the suggestions 👍

PR #427, opened 2025-10-03
handle duplicate context rules

### LINKED ISSUE #393: [M-08] Unlimited and Duplicate Rules in Smart Accounts

brozorec:
Fixes #393 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/accounts/src/smart_account/mod.rs
+++ b/packages/accounts/src/smart_account/mod.rs
@@ -69,8 +69,8 @@ pub trait SmartAccount: CustomAccountInterface {
     ///
     /// # Errors
     ///
-    /// * [`SmartAccountError::TooManyContextRules`] - When the number of context
-    ///   rules exceeds MAX_CONTEXT_RULES (15).
+    /// * [`SmartAccountError::TooManyContextRules`] - When the number of
+    ///   context rules exceeds MAX_CONTEXT_RULES (15).
     /// * [`SmartAccountError::NoSignersAndPolicies`] - When both signers and
     ///   policies are empty.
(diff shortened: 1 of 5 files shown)


packages/accounts/src/smart_account/mod.rs
```diff
@@ -307,7 +311,7 @@ pub enum SmartAccountError {
     /// The specified context rule does not exist.
     ContextRuleNotFound = 2000,
     /// A conflicting context rule already exists.
```
ozgunozerk, line 313: should we also change the error explanation here, and use `duplicate` instead of `conflicting`?

packages/accounts/src/smart_account/test/fingerprints.rs
```diff
@@ -0,0 +1,561 @@
...
+    let contract_addr = Address::generate(&e);
+    let context_type = ContextRuleType::CallContract(contract_addr);
+    let signers = create_test_signers(&e);
+    let policies = Vec::new(&e);
+
+    e.as_contract(&address, || {
+        let fp1 = compute_fingerprint(&e, &context_type, &signers, &policies);
+        let fp2 = compute_fingerprint(&e, &context_type, &signers, &policies);
+        assert_eq!(fp1, fp2); // Should be equal since valid_until is not part
+                              // of fingerprint
```
ozgunozerk, line 153: I think this comment does not make sense here

ozgunozerk (changes requested):
Made small suggestions, great work, will approve as soon as discussions are resolved

ozgunozerk (approved):

PR #432, opened 2025-10-06
recover balance with recovery mechanism

### LINKED ISSUE #377: Missing Same-identity Check in recovery_address Enables Malicious Recovery

ozgunozerk:
Fixes #377 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/src/rwa/identity_verifier/mod.rs
+++ b/packages/tokens/src/rwa/identity_verifier/mod.rs
@@ -53,6 +53,16 @@ pub trait IdentityVerifier {
     ///   identity of the user address cannot be verified.
     fn verify_identity(e: &Env, user_address: &Address);
 
+    /// Returns the target address for the recovery process for the lost wallet.
+    /// If the lost wallet is not a target of a recovery process, `None` is
+    /// returned.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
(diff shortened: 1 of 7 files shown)


packages/tokens/src/rwa/identity_verifier/mod.rs
```diff
@@ -53,6 +53,16 @@ pub trait IdentityVerifier {
     ///   identity of the user address cannot be verified.
     fn verify_identity(e: &Env, user_address: &Address);
```
brozorec, line 54:
```suggestion
   /// * `account` - The user address to verify.
   ///
   /// # Errors
   ///
   /// * [`crate::rwa::RWAError::IdentityVerificationFailed`] - When the
   ///   identity of the user address cannot be verified.
   fn verify_identity(e: &Env, account: &Address);
```

brozorec (approved):

PR #441, opened 2025-10-12
Migrate to target wasmv1-none

### LINKED ISSUE #367: Migrate to `wasmv1-none` from `wasm32-unkown-unkown`

`wasm32v1-none` is relatively new, and the main difference from `wasm32-unknown-unknown` is no `std` support at all (whilst, in `wasm32-unknown-unknown`, there still is some `std` support). 

The problem is, some of our utility crates like `event-assertion` and `default-impl-macro-test` are using `std` code, and with a `cfg` flag, they are not compiled for `wasm32` targets. 

However, since the entire crate is discarded for `wasm32` targets, it seems cargo is filling the empty crate with some `std::prelude` stuff, which is supported by `wasm32-unkown-unkown` but not by `wasm32v1-none`. 

In the future, we may get rid of `default-impl-macro-test` crate for good. 

But, still, we should think of an ergonomic solution for `event-assertion` crate for `wasm32v1-none` target.

jsmaxi:
Fixes #367  

## Migrate target from wasm32-unkown-unkown to wasmv1-none.

This change migrates to `wasm32v1-none` target by mainly updating `test-utils` packages (default-impl-macro-test and event-assertion) to handle `no_std` environments.

Changes:

- Applied `#![cfg_attr(target_family = "wasm", no_std)]` to prevent Cargo from injecting `std::prelude` into empty crates when building for wasm targets that don't support std (like `wasm32v1-none`). Added required panic handler for `no_std` wasm environments.

- Constrained dependencies to non-wasm targets:
```toml
[target.'cfg(not(target_family = "wasm"))'.dependencies]
[target.'cfg(not(target_family = "wasm"))'.dev-dependencies]
```

- Wrapped `event-assertion` std implementation in conditional compilation:
```rust
#[cfg(not(target_family = "wasm"))]
mod implementation {
    /* ... */ 
}
// Re-export for non-wasm targets
#[cfg(not(target_family = "wasm"))]
pub use implementation::*;
```

- Updated readme, architecture, and toolchain information.

The test and build commands continue to function correctly, with the new `wasm32v1-none` target.

#### PR Checklist

- [ ] Tests
- [ ] Documentation

### DIFF
--- a/packages/accounts/src/policies/weighted_threshold.rs
+++ b/packages/accounts/src/policies/weighted_threshold.rs
@@ -128,7 +128,7 @@ pub const WEIGHTED_THRESHOLD_TTL_THRESHOLD: u32 = WEIGHTED_THRESHOLD_EXTEND_AMOU
 /// # Arguments
 ///
 /// * `e` - Access to the Soroban environment.
-/// * `context_rule` - The context rule ID for this policy.
+/// * `context_rule_id` - The context rule ID for this policy.
 /// * `smart_account` - The address of the smart account.
 ///
 /// # Errors
--- a/packages/accounts/src/smart_account/storage.rs
+++ b/packages/accounts/src/smart_account/storage.rs
@@ -12,7 +12,7 @@
 //!   contracts for custom cryptographic verification (e.g., different signature
 //!   schemes)
 //!
-//! ### **What** - Context Rules  
+//! ### **What** - Context Rules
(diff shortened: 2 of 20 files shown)


jsmaxi:
@brozorec Actually good point, after removal, the panic handler becomes unnecessary. I can update the PR

brozorec (approved):

PR #479, opened 2025-10-22
Move vault

ozgunozerk:
- Moves `vault` to root directory for the tokens
- changes error codes for smart accounts and access control to reflect our hierarchy in the codebase


ozgunozerk:
the same uncovered lines (due to max limit for withdraw and deposit) have triggered the codecov patch fail. However, if you take a look at the codecov project, the coverage is increased (I've added 1 more test covering one more line). 

In short, it's good, we can merge

PR #481, opened 2025-10-23
remove antora docs

ozgunozerk:
We migrated docs to another repo, antora docs are no longer required


ozgunozerk:
also disabled the netlify hook

PR #490, opened 2025-11-10
Use constructor args in examples

ElliotFriend:
In a similar thought to OpenZeppelin/contracs-wizard#723, I noticed a lot of the example contract use hard-codes names, symbols, URIs, etc. This PR aims to abstract those variables into constructor function arguments, and provide some helpful comments to document how to take advantage of these arguments at deploy time with the Stellar CLI

#### PR Checklist

Not sure if I'm introducing new features here, but all the tests are passing, and I've updated the various README files with args and comments. If that doesn't count, I'm happy to un-check those boxes lol

- [x] Tests
- [x] Documentation


brozorec (approved):
> I've updated the various README files with args and comments. If that doesn't count, I'm happy to un-check those boxes lol

all good 😄 

thanks @ElliotFriend

ozgunozerk (approved):

PR #494, opened 2025-11-12
timelock

### LINKED ISSUE #436: Timelock Controller
Time-delayed execution of operations for governance and administrative functions (check [GHSA-vrw4-w73r-6mm8](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/security/advisories/GHSA-vrw4-w73r-6mm8) executor privilege escalation vulnerability).

brozorec:
Fixes #436 

In draft in order to validate the approach for self-administrative operations (check the example).

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/examples/timelock-controller/src/contract.rs
+++ b/examples/timelock-controller/src/contract.rs
@@ -88,6 +88,13 @@
 //! This approach ensures administrative changes go through the timelock
 //! process.
 //!
+//! **Note**: Self-administration requires special handling because Soroban does
+//! not allow re-entrancy: a contract cannot call its own public functions
+//! during execution (e.g., `execute_op` cannot internally call `update_delay`
+//! on the same contract). To work around this, the `CustomAccountInterface`
+//! implementation validates and marks operations as executed without performing
+//! the cross-contract call, allowing admin functions to be called directly.
+//!
 //! ## Optional External Admin
 //!
 //! An optional external admin can be provided during deployment to aid with
@@ -107,10 +114,10 @@ use stellar_access::access_control::{
 };
 use stellar_governance::timelock::{
     cancel_operation, execute_operation, get_min_delay as timelock_get_min_delay,
-    get_operation_state, get_timestamp, hash_operation as timelock_hash_operation, is_operation,
-    is_operation_done, is_operation_pending, is_operation_ready, schedule_operation,
-    set_execute_operation, set_min_delay as timelock_set_min_delay, Operation, OperationState,
-    TimelockError,
+    get_operation_state, get_timestamp, hash_operation as timelock_hash_operation,
+    is_operation_done, is_operation_pending, is_operation_ready, operation_exists,
+    schedule_operation, set_execute_operation, set_min_delay as timelock_set_min_delay, Operation,
+    OperationState, TimelockError,
 };
 use stellar_macros::{default_impl, only_admin, only_role};
 
@@ -401,8 +408,8 @@ impl TimelockController {
     }
 
     /// Returns whether an operation exists (scheduled or done).
-    pub fn is_operation(e: &Env, operation_id: BytesN<32>) -> bool {
-        is_operation(e, &operation_id)
+    pub fn operation_exists(e: &Env, operation_id: BytesN<32>) -> bool {
+        operation_exists(e, &operation_id)
     }
 
     /// Returns whether an operation is pending (waiting or ready).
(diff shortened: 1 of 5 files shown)


examples/timelock-controller/src/contract.rs
```diff
@@ -0,0 +1,347 @@
...
+//!
+//! When the contract is deployed with `admin` set to `None`, the contract
+//! address itself becomes the admin, enabling self-administration. For
+//! self-administration operations (e.g., updating the minimum delay):
+//!
+//! 1. Proposer schedules the operation targeting the timelock contract itself
+//! 2. After the delay passes, call the admin function directly (not via
+//!    `execute_op`)
+//! 3. The `CustomAccountInterface` implementation validates the operation is
+//!    ready and marks it as executed
```
ozgunozerk: I'm not familiar with the Solidity one, and this is the first time I'm reading about the subject. And I'm still confused about this `Self-Administration`.

> //! 2. After the delay passes, call the admin function directly (not via `execute_op`)

Who calls? Does the caller need specific privileges, like `admin`? If so, aren't they become the admin in this scenario? Why is it called `Self-Administration`? What does it exactly mean the contract itself become the admin? Then I expect no-one to call the admin functions, but administration functions will be called maybe indirectly throughout other operations with some triggers. 

The explanations here puzzled me...

examples/timelock-controller/src/contract.rs
```diff
@@ -0,0 +1,347 @@
...
+
+#[contractimpl]
+impl TimelockController {
+    /// Initializes the timelock controller.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to Soroban environment.
+    /// * `min_delay` - Initial minimum delay in ledgers for operations.
+    /// * `proposers` - Accounts to be granted proposer and canceller roles.
```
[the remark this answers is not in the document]
brozorec: this example assumes that all initial `proposers` are also `cancelers`, but we keep both roles, because that might change at a later point of contract's life: we can have just `cancelers` without being also `proposers`, e.g. all members can propose and cancel, but we appoint some additional `cancelers` to filter out proposals

examples/timelock-controller/src/contract.rs
```diff
@@ -0,0 +1,347 @@
...
+    /// * `executors` - Accounts to be granted executor role.
+    /// * `admin` - Optional account to be granted admin role for initial setup.
+    ///   If provided, this admin can configure roles without delay but should
+    ///   renounce the role after setup to enforce timelock governance.
+    ///
+    /// # Notes
+    ///
+    /// - The contract itself is always granted the admin role for
+    ///   self-administration.
+    /// - Proposers are automatically granted the canceller role.
```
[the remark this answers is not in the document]
brozorec: cf the comment above

packages/governance/src/timelock/storage.rs
```diff
@@ -0,0 +1,391 @@
...
+    Done,
+}
+
+/// Storage keys for the timelock module.
+#[derive(Clone)]
+#[contracttype]
+pub enum TimelockStorageKey {
+    /// Minimum delay in ledgers for operations
+    MinDelay,
+    /// Maps operation ID to the ledger number when it becomes ready
```
ozgunozerk: > when it becomes ready -> is not clear

packages/governance/src/timelock/storage.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    // Get minimum delay (will panic if not set)
+    let min_delay = get_min_delay(e);
+
+    // Check delay is sufficient
+    if delay < min_delay {
+        panic_with_error!(e, TimelockError::InsufficientDelay);
+    }
+
+    // Calculate ready ledger
```
ozgunozerk: some inline comments are unnecessary, this one is one of them

packages/governance/src/timelock/storage.rs
```diff
@@ -0,0 +1,391 @@
...
+///
+/// **IMPORTANT**: This function does not perform authorization checks.
+/// The caller must ensure proper authorization before calling this function.
+pub fn execute_operation(e: &Env, operation: &Operation) -> Val {
+    set_execute_operation(e, operation);
+
+    e.invoke_contract::<Val>(&operation.target, &operation.function, operation.args.clone())
+}
+
+pub fn set_execute_operation(e: &Env, operation: &Operation) {
```
ozgunozerk: I believe this function is separated for the `check_auth` implementation. 

Inline docs for it are missing. No need to duplicate everything from `execute_operation`, but maybe we can move the inline doc of the `execute operation` here, and then write an inline docs for `execute_operation` from scratch, noting that it is a wrapper which does is extra thing, etc. 

And if we also mention why these two functions are separated, would be good imo

PR #500, opened 2025-11-20
fee forwarder: pay fee in fungible token

### LINKED ISSUE #435: Meta Transactions
- add gas abstraction: use none XLM to pay for gas
- relayer needs to support this gas abstraction

brozorec:
Fixes #435 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/fee-abstraction/src/storage.rs
+++ b/packages/fee-abstraction/src/storage.rs
@@ -112,7 +112,7 @@ pub fn auth_user_and_invoke(
 ///
 /// # Errors
 ///
-/// * refer to [`check_allowed_fee_token`] errors.
+/// * [`FeeAbstractionError::FeeTokenNotAllowed`] - If the token is not allowed.
 /// * refer to [`validate_fee_bounds`] errors.
 #[allow(clippy::too_many_arguments)]
 pub fn collect_fee(
@@ -125,7 +125,9 @@ pub fn collect_fee(
     fee_recipient: &Address,
     approval: FeeAbstractionApproval,
 ) {
-    check_allowed_fee_token(e, fee_token);
+    if !is_allowed_fee_token(e, fee_token) {
+        panic_with_error!(e, FeeAbstractionError::FeeTokenNotAllowed);
+    }
 
     validate_fee_bounds(e, fee_amount, max_fee_amount);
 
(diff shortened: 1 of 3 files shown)


packages/fee-abstraction/src/storage.rs
```diff
@@ -0,0 +1,388 @@
...
+};
+
+// ################## STORAGE KEYS ##################
+
+#[derive(Clone)]
+#[contracttype]
+pub enum FeeAbstractionStorageKey {
+    /// Number of allowed fee tokens
+    Count,
+    /// Allowed fee token, maps to Address
```
ozgunozerk: maps to `Address` of what? Maybe it is worth to mention it here

packages/fee-abstraction/src/storage.rs
```diff
@@ -0,0 +1,388 @@
...
+/// * topics - `["FeeCollected", user: Address, recipient: Address]`
+/// * data - `[token: Address, amount: i128]`
+///
+/// # Errors
+///
+/// * [`FeeAbstractionError::InvalidFeeBounds`] - If amounts <= 0 or `fee_amount
+///   > max_fee_amount`.
+/// * [`FeeAbstractionError::FeeTokenNotAllowed`] - If the fee token is not
+///   allowed when the allowlist is enabled.
+pub fn collect_fee_with_lazy_approval(
```
ozgunozerk: too much code duplication across this one and the eager one. 

For the simplest solution that comes to my mind instantly, there can be an internal helper function that can have an `eager` or `lazy` parameter, and branch on the logic accordingly, and these ones will just call that internal function with the parameter. 

And in that case, one can argue, why not make that internal function the public one and get rid of these two. I'm open for both options (keeping these and having an internal helper, or just having one function with eager/lazy with the parameter). 

I'm just pointing out we should make the code more reusable

packages/fee-abstraction/src/storage.rs
```diff
@@ -0,0 +1,388 @@
...
+///
+/// # Arguments
+///
+/// * `e` - Access to Soroban environment.
+/// * `token` - The token contract address to check.
+///
+/// # Errors
+///
+/// * [`FeeAbstractionError::FeeTokenNotAllowed`] - If the token is not allowed.
+pub fn check_allowed_fee_token(e: &Env, token: &Address) {
```
ozgunozerk: hmm, what is the reason behind not returning a `bool` but directly panicking? From what I've seen, there is no function provided to the users to first check if the token is allowed or not without the panicking behavior.

brozorec: The allow list is enabled only if there's at least one token that was allowed. And this is a convenience function that checks both parts: the allow list is enabled and the provided token is allowed. If it returns a `bool`, it won't be clear which condition is not met.

brozorec: @ozgunozerk that's right, check out now

PR #509, opened 2025-12-02
allowance bug fix

### LINKED ISSUE #507: 🐞 [Bug]: `spend_allowance` may spend expired allowance
### What happened?

`spend_allowance` is calling `allowance_data`, and due to the difference between `allowance` and `allowance_data`, expired allowances may be spent

### Expected behavior

_No response_

ozgunozerk:
Fixes #507 

#### PR Checklist

- [x] Tests
- [x] Documentation

Notes for @brozorec :
- `allowance` has this expiry check, however, we also need to access the `live_until_ledger` info from `allowance_data` in the `spend_allowance` function. That's why, simply using `allowance` instead of `allowance_data` in `spend_allowance` wouldn't work.
- `allowance_data` is not checking for the expiry, and delegates that responsibility to the caller function. Considering that we ourselves made the mistake of using `allowance_data` but not enforcing the expiry check (and this has not been caught by the previous audits), I thought we better change the design of the `allowance_data`, and put the EXPIRY CHECK logic in it. 
- the original design of the `allowance_data` was not panicking. Even if the allowance is not found, it was returning an allowance data object with defaults (`amount: 0, live_until_ledger: 0`).
- what to return when it is expired? I first thought, (`amount: 0, live_until_ledger: allowance_data.live_until_ledger`). Because, maybe the information of when the allowance has expired will be important. But, then, the same argument can be made for the `amount`... 
- In the light of the above arguments, I've changed the behavior of `allowance_data` to include panicking. It will panic when an allowance is not found, and it will panic when an allowance has expired, which allowed me to make further simplifications across the codebase, such as:
    - `amount == 0` was both a valid allowance, and also signaling expired allowances. This resulted in hard to decipher logic in some parts of the code
    - now, `amount == 0` is a valid allowance, but not signaling expired allowances. Hence, an allowance with `amount: 0, live_until_ledger: EXPIRED_LEDGER` is not valid to set anymore. And **it was valid to set before**.

I think all these changes made the library simpler and more secure. Let me know if you want to further discuss on them!

### DIFF
--- a/packages/tokens/src/fungible/extensions/burnable/test.rs
+++ b/packages/tokens/src/fungible/extensions/burnable/test.rs
@@ -66,7 +66,7 @@ fn burn_with_insufficient_balance_panics() {
 }
 
 #[test]
-#[should_panic(expected = "Error(Contract, #101)")]
+#[should_panic(expected = "Error(Contract, #116)")]
 fn burn_with_no_allowance_panics() {
     let e = Env::default();
     e.mock_all_auths();
--- a/packages/tokens/src/fungible/mod.rs
+++ b/packages/tokens/src/fungible/mod.rs
@@ -310,10 +310,14 @@ pub enum FungibleTokenError {
     SACMissingFnParam = 111,
     /// Indicates an invalid function parameter in the SAC contract context.
     SACInvalidFnParam = 112,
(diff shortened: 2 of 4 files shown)


ozgunozerk:
closing in favor of #513

PR #512, opened 2025-12-03
smart account fixes

brozorec:
This PR:
- adds a new read function to the interface: `get_context_rules_count`
- fixes the off-by-one error in the spending limit policy’s period. When [older spending entries are pruned](https://github.com/OpenZeppelin/stellar-contracts/blob/372a162a6df8aed108befa5573d30a8ca49cb271/packages/accounts/src/policies/spending_limit.rs#L433), the cutoff is current_ledger - period, which effectively allows period + 1 ledgers instead of period, potentially causing discrepancies
- prevent calling install on available policies for already installed smart accounts and context rules
- modifies the webauthn verifier in examples to accommodate storing of the credential id next to public key

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/examples/multisig-smart-account/README.md
+++ b/examples/multisig-smart-account/README.md
@@ -58,7 +58,7 @@ stellar contract deploy --alias webauthn_verifier \
 
 For this example, we assume the passkey verifier was deployed to:
 ```
-CC74I4FO7N2VY547Q7HA4QFIBZTYVG76DJH32ICBKI2FUPAS6J6MJE66
+CBEO6Q7UXBIQIHQR42RXETMYDKW7GABRX2O4UVW6O6YQOHROYWZJCOXZ
 ```
 
 ### Deploy Ed25519 Verifier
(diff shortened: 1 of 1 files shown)


examples/multisig-smart-account/webauthn-verifier/src/contract.rs
```diff
@@ -39,9 +45,13 @@ impl Verifier for WebauthnVerifierContract {
     ///
     /// * `true` if the signature is valid
     /// * `false` otherwise
-    fn verify(e: &Env, signature_payload: Bytes, key_data: BytesN<65>, sig_data: Bytes) -> bool {
+    fn verify(e: &Env, signature_payload: Bytes, key_data: Bytes, sig_data: Bytes) -> bool {
```
ozgunozerk: I see that `verify` function is using `Self::KeyData` for the type of `key_data`. I think it is better use it like that also in here, because otherwise, providing the associated type seems pointless.

PR #523, opened 2025-12-19
vault consistency fix

### LINKED ISSUE #524: Consistency for Vault implementation
In all of our other tokens, the end developer does not need to think about `require_auth()` for the default implementations, but this is not the case for vault. We should put `require_auth()` for the default implementations for vault as well.

ozgunozerk:
Fixes #524 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/packages/tokens/src/vault/storage.rs
+++ b/packages/tokens/src/vault/storage.rs
@@ -326,14 +326,9 @@ impl Vault {
     ///   Address]`
     /// * data - `[assets: i128, shares: i128]`
     ///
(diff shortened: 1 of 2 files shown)


packages/tokens/src/vault/storage.rs
brozorec: replace with a mention that `operator` authorization is required

brozorec: replace with a mention that `operator` authorization is required

PR #525, opened 2025-12-19
Contract trait

### LINKED ISSUE #526: 🎁 [Feature Request]: deprecate `#[default_impl]` and use the new `#[contractimpl(contracttrait)]` macro
### What is the feature you would like to see?

Due to the limitations of the rust macro scopes, we had to find a workaround (which was `#[default_impl]` to solve the problem of bring the default implementations to the scope of actual smart contract implementation.

```rust
/// `#[contractimpl]` macro requires all the default implementations to be
/// provided under the code block:
///
/// ```ignore
/// #[contractimpl]
/// impl Trait for MyContract {
///     /*
///         The client generated by the `#[contractimpl]` macro will have access only
///         to the methods specified in here. Which means, if you do not provide the
///         default implementations of the methods, the client generated by the
///         `#[contractimpl]` macro won't have access to those methods.
///
///         This is due to how macros work in Rust. They cannot access the default
///         implementations of the methods of this trait, since they are not in the
///         scope of the macro.
///
///         To overcome this, we provide a macro for our traits, that generates the
///         missing default implementations for the methods of the trait, so
///         you can only focus on the overrides, and leave the default implementations
///         out as per usual.
///     */
/// }
/// ```
///
/// # Example:
///
/// ```ignore
/// #[default_impl] // IMPORTANT: place this above `#[contractimpl]`
/// #[contractimpl]
/// impl NonFungibleToken for MyContract {
///     /* your overrides here */
/// }
/// ```
```

However, with some ingenious workarounds, now `#[contractimpl]` macro overcomes this obstacle. 

Here is the related [issue](https://github.com/stellar/rs-soroban-sdk/issues/1451) and [PR](https://github.com/stellar/rs-soroban-sdk/pull/1507) on the soroban sdk repo.

If you are still curious on how it works, here is an [article](https://cryptical.xyz/rust/shortcomings-of-macros).

ozgunozerk:
Fixes #526 

#### PR Checklist

- [x] Tests
- [x] Documentation

### DIFF
--- a/examples/nft-sequential-minting/src/test.rs
+++ b/examples/nft-sequential-minting/src/test.rs
@@ -37,3 +37,17 @@ fn burn_works() {
     client.burn(&owner, &0);
     assert_eq!(client.balance(&owner), 0);
 }
+
+#[test]
+fn burn_from_works() {
+    let e = Env::default();
+    let owner = Address::generate(&e);
+    let spender = Address::generate(&e);
+    let client = create_client(&e, &owner);
+
+    e.mock_all_auths();
+    client.mint(&owner);
+    client.approve(&owner, &spender, &0, &1000);
+    client.burn_from(&spender, &owner, &0);
+    assert_eq!(client.balance(&owner), 0);
+}


brozorec (approved):
That's a massive improvement, great job 👏 👏 👏 

Should we also provide defaults for `allowed()` and `blocked()` in the fungible extensions?

PR #544, opened 2025-12-29
[L-02] - refactor fixed point lib

### LINKED ISSUE #532: [L-02] - Wad::checked_mul and Wad::pow Rounding Behavior Disagrees With Documentation for Negative Results

ozgunozerk:
Fixes #532 

#### PR Checklist

- [x] Tests
- [x] Documentation

Notes: this issue fix required us to expose a new variant of `mul_div` without any rounding, because that functionality is used in a lot of places in `WAD`, and having this function as a helper in `WAD` does make less sense then having it in our fixed point library. 

The best place to include that, was in the trait. Then, I've realized some inconsistencies between I256 and i128 libraries. 

In short, changing the trait, and fixing the inconsistencies, led me to refactoring the fixed point library altogether.

Here are the summary of the changes:
- trait implementations were not done in place, but were delegated to other functions. This led to naming problems, and also the helper functions were not separated properly. Additionally, this approach is not inline with other parts of our library (where we have the trait, and the storage files separately). So, I moved the implementations directly into the traits for i128 and i256
- added 2 new methods to the traits, truncate variant for both checked and unchecked `mul_div`
- added a variant to `Rounding` enum, which is `Truncate`
- added `muldiv` and `checked_muldiv` standalone functions to `i256` as well, because, any difference across `i128` and `i256` is confusing
- the way we implement `checked` and `non_checked` functions in `i256` was not consistent with `i128`, changed them as well
- removed the `fixed_point.rs` test, because we don't have any function/method exposed from the `mod.rs` file anymore
- renamed the trait to `SorobanMulDiv`, (previously `SorobanFixedPoint`), because we are not providing a suite of fixed point arithmetic math under that trait, but only the variants of `mul_div`

### DIFF
--- a/packages/contract-utils/src/math/i128_fixed_point.rs
+++ b/packages/contract-utils/src/math/i128_fixed_point.rs
@@ -76,28 +76,28 @@ impl SorobanMulDiv for i128 {
     ///
     /// # Errors
     ///
-    /// Returns [`SorobanFixedPointError::DivisionByZero`] if denominator is
-    /// zero. Returns [`SorobanFixedPointError::Overflow`] if the result
-    /// overflows.
-    fn mul_div_floor(&self, env: &Env, y: &i128, denominator: &i128) -> i128 {
+    /// * [`SorobanFixedPointError::DivisionByZero`] - when `denominator` is
+    ///   zero.
+    /// * [`SorobanFixedPointError::Overflow`] - when the result overflows.
+    fn mul_div_floor(&self, e: &Env, y: &i128, denominator: &i128) -> i128 {
         if *denominator == 0 {
-            panic_with_error!(env, SorobanFixedPointError::DivisionByZero);
+            panic_with_error!(e, SorobanFixedPointError::DivisionByZero);
         }
         match self.checked_mul(*y) {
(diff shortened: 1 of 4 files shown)


ozgunozerk:
I'll write additional tests for coverage, and make it 100% for this PR

packages/contract-utils/src/math/i128_fixed_point.rs
```diff
@@ -49,19 +50,207 @@ pub fn muldiv(e: &Env, x: i128, y: i128, denominator: i128, rounding: Rounding)
...
+    /// * `env` - Access to the Soroban environment.
+    /// * `y` - The multiplicand.
+    /// * `denominator` - The divisor.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`SorobanFixedPointError::DivisionByZero`] if denominator is
+    /// zero. Returns [`SorobanFixedPointError::Overflow`] if the result
+    /// overflows.
+    fn mul_div_floor(&self, env: &Env, y: &i128, denominator: &i128) -> i128 {
```
brozorec: use `e` instead of `env` for consistency here and below

brozorec (approved):
aside from some naming and docstring inconsistencies, great job 🙌

PR #561, opened 2026-02-02
Negative trait bounds

### LINKED ISSUE #560: 🏗️ [Core Feature]: Have a better architecture to enforce mutual exclusivity, without limiting composability and worsening DevX

- a contract with A and C

- AandC
- AandD
- BandC
- BandD

ozgunozerk:
Fixes #560 

#### PR Checklist

- [x] Tests
- [x] Documentation

P.S. strongly suggest to run `cargo clean` on your workspace after this merges, because we are switching toolchain. This will prevent non-sense weird error messages. Speaking from experience :') 

## Summary by CodeRabbit

* **Documentation**
  * Revised architecture documentation to reflect override-based extension mechanism with compile-time constraint enforcement.

* **New Features**
  * Added allowlist trait methods: `allowed()`, `allow_user()`, and `disallow_user()` for managing permitted accounts.

* **Refactor**
  * Simplified trait inheritance requirements for token extensions.
  * Established mutual exclusivity between conflicting extension implementations (allowlist/blocklist, enumerable/consecutive).

### DIFF
--- a/.github/workflows/generic.yml
+++ b/.github/workflows/generic.yml
@@ -47,7 +47,7 @@ jobs:
       - name: install rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: stable, nightly
+          toolchain: nightly
           components: clippy, rustfmt, llvm-tools-preview
           target: wasm32v1-none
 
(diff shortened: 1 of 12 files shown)


ozgunozerk:
On top of that, cargo automatically handles the change behind the scenes. Having some projects using nightly, and some using stable, at the same time, is not that uncommon in rust. The end developers may not even notice cargo is using nightly in the background

ozgunozerk:
that's weird. Because I didn't put any command on my workflow to change the toolchain. I just updated the toolchain file in the repo, and voila!

Let's go to slack for discussing this further

ozgunozerk:
closing in favor of #821

PR #563, opened 2026-02-06
Governance base contract

### LINKED ISSUE #556: Governor
- [ ] Governor Base Contract
- Extensions:
    - [ ] GovernorCountingSimple
    - [x] GovernorVotes
    - [ ] GovernorSettings
    - [x] GovernorTimelockControl

ozgunozerk:
Fixes #556 

#### PR Checklist

- [ ] Tests // TODO: after Votes and Counting are merged
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive on-chain governance system with proposal creation, voting, execution, and cancellation capabilities.
  * Added governance events for proposal lifecycle tracking (created, voted, queued, executed, cancelled).
  * Added proposal state management and voting period configuration.

* **Tests**
  * Added test scaffolding for governance module.

### DIFF
--- a/packages/tokens/src/non_fungible/storage.rs
+++ b/packages/tokens/src/non_fungible/storage.rs
@@ -477,7 +477,9 @@ impl Base {
             return;
         }
 
-        if live_until_ledger < e.ledger().sequence() {
+        if live_until_ledger > e.ledger().max_live_until_ledger()
+            || live_until_ledger < e.ledger().sequence()
+        {
             panic_with_error!(e, NonFungibleTokenError::InvalidLiveUntilLedger);
         }
 
--- a/packages/tokens/src/rwa/identity_registry_storage/storage.rs
+++ b/packages/tokens/src/rwa/identity_registry_storage/storage.rs
@@ -410,10 +410,7 @@ pub fn add_identity(
 pub fn modify_identity(e: &Env, account: &Address, new_identity: &Address) {
     let key = IRSStorageKey::Identity(account.clone());
 
-    let old_identity: Address = e
-        .storage()
-        .persistent()
-        .get(&key)
+    let old_identity = get_persistent_entry(e, &key)
         .unwrap_or_else(|| panic_with_error!(e, IRSError::IdentityNotFound));
 
     e.storage().persistent().set(&key, new_identity);
@@ -523,10 +520,7 @@ pub fn recover_identity(e: &Env, old_account: &Address, new_account: &Address) {
     let old_identity_key = IRSStorageKey::Identity(old_account.clone());
     let new_identity_key = IRSStorageKey::Identity(new_account.clone());
 
-    let identity: Address = e
-        .storage()
-        .persistent()
-        .get(&old_identity_key)
+    let identity: Address = get_persistent_entry(e, &old_identity_key)
         .unwrap_or_else(|| panic_with_error!(e, IRSError::IdentityNotFound));
 
     // Check if new_account is not already linked to another identity
--- a/packages/tokens/src/rwa/storage.rs
+++ b/packages/tokens/src/rwa/storage.rs
@@ -706,6 +706,8 @@ impl RWA {
     pub fn transfer_from(e: &Env, spender: &Address, from: &Address, to: &Address, amount: i128) {
         spender.require_auth();
 
+        Self::validate_transfer(e, from, to, amount);
+
         Base::spend_allowance(e, from, spender, amount);
 
         Base::update(e, Some(from), Some(to), amount);


packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+/// Base Governor Trait
+///
+/// The `Governor` trait defines the core functionality for on-chain governance.
+/// It provides a standard interface for creating proposals, voting, and
+/// executing approved actions.
+///
+/// The contract that implements this trait is expected to implement [`Votes`]
+/// and [`Counting`] traits.
+#[contracttrait]
+pub trait Governor: Votes + Counting {
```
brozorec: I find confusing that authorizations are handled differently for each function:
- it's missing for `execute()`
- from the storage functions for `propose()` and `cancel()` with quite opinionated approach for `cancel()` assuming that proposers are also cancellers
- from some other place for `cast_vote()`

packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+    ///
+    /// # Events
+    ///
+    /// * topics - `["proposal_created", proposal_id: BytesN<32>, proposer:
+    ///   Address]`
+    /// * data - `[targets: Vec<Address>, functions: Vec<Symbol>, args:
+    ///   Vec<Vec<Val>>, vote_start: u32, vote_end: u32, description: String]`
+    fn propose(
+        e: &Env,
+        proposer: Address,
```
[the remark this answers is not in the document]
brozorec: if you are ok with this, move those params in the corresponding functions as well

packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+    /// The proposal contains no actions.
+    EmptyProposal = 5003,
+    /// The targets, functions, and args vectors have different lengths.
+    InvalidProposalLength = 5004,
+    /// The proposal is not in the active state.
+    ProposalNotActive = 5005,
+    /// The proposal has not succeeded.
+    ProposalNotSuccessful = 5006,
+    /// The proposal has not been queued.
+    ProposalNotQueued = 5007,
```
brozorec: unused

packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,683 @@
...
+    }
+}
+
+// ################## TYPES ##################
+
+/// The state of a proposal in its lifecycle.
+#[contracttype]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum ProposalState {
```
brozorec: It looks straightforward at 1st sight, but as I was going through the code, I had some difficulties in reasoning about the subcategories of those states, like terminal and non-terminal states, being set from the governor vs extensions. Taking into account those specifics will be helpful in avoiding bugs.

packages/governance/src/governor/storage.rs
```diff
@@ -0,0 +1,611 @@
...
+    // Get proposal and verify it exists
+    let mut proposal = get_proposal_core(e, &proposal_id);
+
+    // Only proposer can cancel
+    proposal.proposer.require_auth();
+
+    // Can only cancel proposals that haven't reached a terminal state
+    let state = get_proposal_state(e, &proposal_id);
+    match state {
+        ProposalState::Pending | ProposalState::Active | ProposalState::Succeeded => {}
```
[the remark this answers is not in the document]
ozgunozerk: I think yes, I don't see a reason why not? 

The proposer is suggesting something, and unless executed, he can cancel the proposal. If the community wants it (say, it got a lot of votes), someone else can propose it again. I think it is useful to have this flexibility.

Maybe, the proposer wants to do smt that will mostly benefit himself, and others don't mind, so they voted yes. But the circumstances have changed, and now it is not beneficial for the proposer, and others still don't mind.

packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,763 @@
...
+use soroban_sdk::{
+    contracterror, contractevent, contracttrait, contracttype, Address, BytesN, Env, String,
+    Symbol, Val, Vec,
+};
+
+/// TODO: delete this after Votes PR is merged
+pub trait Votes {
+    fn get_past_votes(e: &Env, account: Address, ledger: u32) -> u128;
+    fn get_past_total_supply(e: &Env, ledger: u32) -> u128;
+}
```
brozorec: `fn get_past_votes` corresponds now to  `fn get_votes_at_checkpoint(e: &Env, account: Address, timepoint: u64) -> u128;` and `fn get_past_total_supply` to `fn get_total_supply_at_checkpoint(e: &Env, timepoint: u64) -> u128;`

packages/governance/src/governor/storage.rs
```diff
@@ -0,0 +1,639 @@
...
+/// * `args` - The arguments for each function call.
+/// * `description_hash` - The hash of the proposal description.
+pub fn hash_proposal(
+    e: &Env,
+    targets: &Vec<Address>,
+    functions: &Vec<Symbol>,
+    args: &Vec<Vec<Val>>,
+    description_hash: &BytesN<32>,
+) -> BytesN<32> {
+    use soroban_sdk::Bytes;
```
brozorec: Is there any specific reason for not moving it on top-level?

packages/governance/src/governor/storage.rs
```diff
@@ -0,0 +1,639 @@
...
+///   the active state.
+/// * refer to [`get_proposal_state()`] errors.
+pub fn prepare_vote(e: &Env, proposal_id: &BytesN<32>) -> u32 {
+    let state = get_proposal_state(e, proposal_id);
+    if state != ProposalState::Active {
+        panic_with_error!(e, GovernorError::ProposalNotActive);
+    }
+
+    get_proposal_snapshot(e, proposal_id)
+}
```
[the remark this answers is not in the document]
ozgunozerk: good catch!

packages/governance/src/governor/mod.rs
```diff
@@ -0,0 +1,843 @@
...
+///   set by extensions like `TimelockControl`.
+/// - [`Executed`](ProposalState::Executed) — set by the Governor.
+#[contracttype]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum ProposalState {
+    // ==================== Time-based states ====================
+    // Derived from the current ledger.
+    /// The proposal is pending and voting has not started yet.
+    Pending = 0,
```
brozorec: We actually store the initial state as `Pending` so the docs here are a bit misleading.

packages/governance/src/governor/storage.rs
```diff
@@ -0,0 +1,639 @@
...
+        }
+        _ => {}
+    }
+
+    // Mark as cancelled
+    proposal.state = ProposalState::Canceled;
+    e.storage().persistent().set(&GovernorStorageKey::Proposal(proposal_id.clone()), &proposal);
+
+    // Emit event
+    crate::governor::emit_proposal_cancelled(e, &proposal_id);
```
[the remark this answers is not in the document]
brozorec: Unresolving this to nuance the point we had about full paths.

I'm fine with the argument to use full paths for the default implementation in `mod.rs` like `storage::my_func()` to avoid possible confusions, but here I think it's an overkill. Furthermore, if we want to change this practice across the library, we should do it in a more consistent way, ie. once for all, and not on a module-per-module basis.

brozorec (changes requested):
Great re-work 🙌 I think the flow looks more straightforward now. I have couple more suggestions.

brozorec (approved):
Awesome 🙌  only couple of nit picks and it's ready to go.

PR #581, opened 2026-02-23
Smart accounts: key canonicalization, size limits and direct rule hints

### LINKED ISSUE #570: smart accounts: Enumerating Key Representations instead of External Signer Identities
Since only the verifier knows the mapping from signer bytes to cryptographic identity, external verifiers should implement a function that returns a canonical identity representation for a given key. The Smart Account will then use this value to prevent duplicates at registration time.

### LINKED ISSUE #571: smart accounts: add max size limit to external signers
One oversized rule breaks get_context_rules() for all rules of that context type so we need to enforce maximum key data size at creation.

### LINKED ISSUE #572: smart accounts: select a specific context rule to match against during auth
If the user knows what rule they want to auth against, there's little interest to make them iterate over all previously added context rules. The current structure can get very expensive and entirely needlessly so. This would allow alleviating the 15 context rule limitation as well.

brozorec:
Fixes #570
Fixes #571
Fixes #572 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added direct rule selection for context rules via hints to bypass standard iteration
  * Implemented key canonicalization for Ed25519 and WebAuthn verifiers
  * Added external signer key size validation
  * Introduced new `get_context_rule_ids` API for efficient rule ID retrieval

* **Breaking Changes**
  * Signatures structure updated from tuple to named-field struct
  * `get_context_rules` deprecated; use `get_context_rule_ids` instead

* **Documentation**
  * Updated authorization flow documentation with new rule selection scenarios and migration guide

### DIFF
--- a/packages/accounts/src/policies/test/spending_limit.rs
+++ b/packages/accounts/src/policies/test/spending_limit.rs
@@ -656,3 +656,129 @@ fn history_capacity_allows_new_transaction_after_cleanup() {
         enforce(&e, &context, &context_rule.signers, &context_rule, &smart_account);
     });
 }
+
+// These tests exercise the error paths in `enforce` that require non-empty
+// `authenticated_signers` so the early empty-signer check is bypassed and
+// execution reaches the targeted code path.
+
+#[test]
+#[should_panic(expected = "Error(Contract, #3223)")]
+fn enforce_non_transfer_fn_name_with_signers_errors() {
+    let e = Env::default();
+    let address = e.register(MockContract, ());
+    let smart_account = Address::generate(&e);
+    let context_rule = create_context_rule(&e);
+
+    e.mock_all_auths();
+
+    e.as_contract(&address, || {
+        let params = SpendingLimitAccountParams { spending_limit: 1_000_000, period_ledgers: 100 };
+        install(&e, &params, &context_rule, &smart_account);
+    });
+
+    e.as_contract(&address, || {
+        let contract_address = Address::generate(&e);
+        let context = Context::Contract(ContractContext {
+            contract: contract_address,
+            fn_name: symbol_short!("deploy"),
+            args: Vec::new(&e),
+        });
+
+        enforce(&e, &context, &context_rule.signers, &context_rule, &smart_account);
+    });
+}
+
+#[test]
+#[should_panic(expected = "Error(Contract, #3223)")]
+fn enforce_missing_amount_arg_with_signers_errors() {
+    let e = Env::default();
+    let address = e.register(MockContract, ());
+    let smart_account = Address::generate(&e);
+    let context_rule = create_context_rule(&e);
+
+    e.mock_all_auths();
+
+    e.as_contract(&address, || {
+        let params = SpendingLimitAccountParams { spending_limit: 1_000_000, period_ledgers: 100 };
+        install(&e, &params, &context_rule, &smart_account);
+    });
+
+    e.as_contract(&address, || {
+        let contract_address = Address::generate(&e);
+        let from = Address::generate(&e);
+        let to = Address::generate(&e);
+
+        let mut args = Vec::new(&e);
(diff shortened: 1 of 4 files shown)


packages/accounts/README.md
```diff
@@ -203,6 +205,32 @@ For each rule in order:
...
+
+```rust
+// Two auth contexts → two rule IDs, one per context.
+Signatures {
+    signers: signature_map,
+    context_rule_ids: vec![&e, rule_a_id, rule_b_id],
+}
+```
+
+The length of `context_rule_ids` must equal the number of auth contexts; a mismatch is rejected with `ContextRuleIdsLengthMismatch`. When the vec is empty, the standard newest-first iteration applies to every context.
```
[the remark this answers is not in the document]
ozgunozerk: that was also in some other place, so search throughout the repo would be good if you want to change this

packages/accounts/README.md
```diff
@@ -389,13 +433,84 @@ For external signers, there are two options:
...
-- Multiple context rules for the same context can co‑exist; the most recently added one takes precedence.
+- Multiple context rules for the same context can co‑exist; the most recently added one takes precedence when using the default iteration algorithm.
+- When using `context_rule_ids` to target rules directly, iteration order is bypassed entirely. Each specified rule is validated on its own merits regardless of whether newer rules exist for the same context. Remove outdated permissive rules when you want to enforce stricter ones unconditionally.
 - For simple cases like threshold‑based multisig, using a policy may feel verbose compared to embedding logic directly in the account, but keeping business rules in policies preserves separation of concerns and allows for a greater flexibility.
 - Authorization composes independent contracts (the smart account, verifiers, and policies). Protocol 23 makes cross‑contract calls cheap, but not free, so the framework sets explicit limits to keep costs predictable:
   - Maximum signers per context rule: 15
   - Maximum policies per context rule: 5
   - Maximum context rules per smart account: 15
 
+## Migration Guide: from v0.6.0 to 0.7.0
```
ozgunozerk: I think this is better put in release notes than here. Definitely valuable, but wrong place imo

ozgunozerk (changes requested):
in `smart_account/storage.rs`, the `# Errors` section should refer to the relevant function calls instead of unpacking all the errors

packages/accounts/src/smart_account/storage.rs
```diff
@@ -804,41 +904,52 @@ pub fn update_context_rule_valid_until(e: &Env, id: u32, valid_until: Option<u32
...
+        signers: resolve_signers(e, &entry.signer_ids),
+        policies: policies.clone(),
+        valid_until: entry.valid_until,
+    };
 
-    // Uninstall all policies
-    for policy in context_rule.policies.iter() {
-        // try_uninstall so that if the policy panics, complete the removal
+    for (policy, policy_id) in policies.iter().zip(&entry.policy_ids) {
+        // try_uninstall so that if the policy panics, context rule removal can be
```
ozgunozerk: can you elaborate on why do we want `context_rule_removal` to succeed when `policy_uninstall` fails?

packages/accounts/src/smart_account/storage.rs
```diff
@@ -1095,26 +1496,58 @@ pub fn validate_and_set_fingerprint(
...
 }
 
+/// Resolves a list of global signer IDs to their full [`Signer`] objects by
+/// reading each [`SignerEntry`] from persistent storage.
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `signer_ids` - The global signer IDs to resolve.
+fn resolve_signers(e: &Env, signer_ids: &Vec<u32>) -> Vec<Signer> {
```
ozgunozerk: same for here, I think `get_signer_addresses` or some other naming would be better instead of `resolve_signer`.

Issue #562, opened 2026-02-05
🐞 [Bug]: UpgradeableMigratable macro leaks type alias into contract spec, breaking migrate invocation


fazzatti:
### What happened?

The `UpgradeableMigratable` derive macro generates a contract interface that references a type alias (`MigrationData`) which does not exist in the Soroban contract spec.

When `UpgradeableMigratableInternal` is implemented with `type MigrationData = ()`, the macro exposes `MigrationData` as a named parameter type in the `migrate` function. Type aliases are not supported by the rs-soroban-sdk spec generator, so the generated spec contains a reference to `MigrationData` without any definition.

As a result, spec driven tooling (JS SDK, generated bindings) cannot encode the `migrate` arguments. Attempting to invoke the function manually using `ScVal::Void` also fails at runtime with:

```
HostError: Error(WasmVm, InvalidAction)
VM call trapped: UnreachableCodeReached
```

This is not an issue with using `()` as migration data, but with the macro leaking a Rust-only alias into the public contract ABI. Ideally, the macro should expose a spec supported type directly instead of a named alias.

The issue can be reproduced with the following contract implementation:

```
use soroban_sdk::{contract, contractimpl, Address, Env};
use stellar_contract_utils::upgradeable::UpgradeableMigratableInternal;
use stellar_macros::UpgradeableMigratable;

#[derive(UpgradeableMigratable)]
#[contract]
pub struct UpgradeableContract;

#[contractimpl]
impl UpgradeableContract {}

impl UpgradeableMigratableInternal for UpgradeableContract {
    type MigrationData = ();
    fn _migrate(_env: &Env, _migration_data: &Self::MigrationData) {
        // No migration logic
    }

    fn _require_auth(_e: &Env, operator: &Address) {
        operator.require_auth();
    }
}
```

Invoking `migrate` by passing `ScVal::Void` for `migration_data` fails at runtime with a Wasm trap (`Error(WasmVm, InvalidAction)`).

For convenience, here is a transaction XDR that reproduces the failure when simulated:

[...]

PR #585, opened 2026-02-24
Remove Upgradeable derive macros 

### LINKED ISSUE #519: Migration Fails with `Missing Entry MigrationData` When Using Custom Migration Data name
### Summary  
When testing the upgradeable contract example on **Stellar testnet**, the migration step fails with:
`error: Missing Entry MigrationData`

This happens **only when the migration data struct is not literally named `MigrationData`**, even though the example code suggests that the struct name should not matter.

### Steps to Reproduce

Using the example upgradeable contracts:

```bash
#!/bin/bash
# Build and deploy V1 contract
cargo build --release --target wasm32-unknown-unknown -p upgradeable-v1-example
stellar contract deploy --wasm target/wasm32-unknown-unknown/release/upgradeable_v1_example.wasm \
  --source alice --network testnet --alias upgradeable_v1 -- --admin $ALICE_ADDRESS

# Build and upload V2 contract
cargo build --release --target wasm32-unknown-unknown -p upgradeable-v2-example
V2_HASH=$(stellar contract upload \
  --wasm target/wasm32-unknown-unknown/release/upgradeable_v2_example.wasm \
  --source alice --network testnet)

# Upgrade V1 to V2
stellar contract invoke --id upgradeable_v1 --source alice --network testnet --send yes \
  -- upgrade --new_wasm_hash $V2_HASH --operator $ALICE_ADDRESS

# Attempt migration (THIS FAILS)
stellar contract invoke --id upgradeable_v1 --source alice --network testnet --send yes \
  -- migrate --migration_data '{"num1": 1000, "num2": 2000}' --operator $ALICE_ADDRESS
```

Actual Result：

❌ error: Missing Entry MigrationData

### Additional Observations

If I rename my migration data struct exactly to MigrationData, the migration succeeds.

```
mod newtypes{
    use soroban_sdk::contracttype;

    #[contracttype]
    pub struct MigrationData {
        pub num1: u32,
        pub num2: u32,
    }
}

use newtypes::MigrationData as Data;

```

If I use any other struct name (e.g., MyMigrationData), the migration fails with the above error.

The example code in the repository indicates that struct naming should not be important for migration, but this does not match observed behavior on testnet.

### Questions

Is this a bug in the example or in the underlying CLI/runtime?

Is the struct name MigrationData currently hard-coded or required in some part of the upgrade flow?

brozorec:
## Summary

Replaces the `#[derive(Upgradeable)]` and `#[derive(UpgradeableMigratable)]` derive macros — along with the three internal traits they relied on — with two plain helper functions: `upgrade()` and `run_migration()`. Users now implement the `Upgradeable` trait directly.

Fixes #519 and #562
The old `UpgradeableMigratable` derive macro generated a module-level `type MigrationData = ...` alias that leaked into the contract WASM spec as an undefined type, breaking CLI invocation and JS SDK bindings. With explicit impls the concrete type (e.g. `Data`) appears directly in the function signature.

### Before

```rust
#[derive(Upgradeable)]
#[contract]
pub struct ExampleContract;

impl UpgradeableInternal for ExampleContract {
    fn _require_auth(e: &Env, operator: &Address) { ... }
}
```

### After

```rust
#[contract]
pub struct ExampleContract;

#[contractimpl]
impl Upgradeable for ExampleContract {
    fn upgrade(e: &Env, new_wasm_hash: BytesN<32>, operator: Address) {
        // access control here
        upgradeable::upgrade(e, &new_wasm_hash);
    }
}
```

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed automatic upgrade code generation; contracts must now explicitly implement upgrade and migration logic.
  * Upgrade authorization changed from owner-based checks to role-based access control.
  * Constructor initialization now uses role-based admin setup instead of manual ownership tracking.

* **New Features**
  * Explicit upgrade and migration functions with configurable role-based authorization.

### DIFF
--- a/examples/nft-access-control/src/test.rs
+++ b/examples/nft-access-control/src/test.rs
@@ -486,7 +486,7 @@ fn burners_can_call_multi_role_action() {
 }
 
 #[test]
-#[should_panic(expected = "Account does not have any of the required roles")]
+#[should_panic(expected = "Error(Contract, #2000)")]
 fn outsiders_cannot_call_multi_role_action() {
     let e = Env::default();
     let admin = Address::generate(&e);
@@ -531,7 +531,7 @@ fn burners_can_call_multi_role_auth_action() {
 }
 
 #[test]
-#[should_panic(expected = "Account does not have any of the required roles")]
+#[should_panic(expected = "Error(Contract, #2000)")]
 fn outsiders_cannot_call_multi_role_auth_action() {
     let e = Env::default();
     let admin = Address::generate(&e);
--- a/packages/access/src/access_control/storage.rs
+++ b/packages/access/src/access_control/storage.rs
@@ -6,7 +6,7 @@ use crate::{
         emit_role_admin_changed, emit_role_granted, emit_role_revoked, AccessControlError,
         MAX_ROLES, ROLE_EXTEND_AMOUNT, ROLE_TTL_THRESHOLD,
     },
-    role_transfer::{accept_transfer, transfer_role},
+    role_transfer::{accept_transfer, transfer_role, PendingTransfer},
 };
 
 /// Storage key for enumeration of accounts per role.
@@ -463,7 +463,7 @@ pub fn renounce_admin(e: &Env) {
     let admin = enforce_admin_auth(e);
     let key = AccessControlStorageKey::PendingAdmin;
 
-    if e.storage().temporary().get::<_, Address>(&key).is_some() {
+    if e.storage().temporary().get::<_, PendingTransfer>(&key).is_some() {
         panic_with_error!(e, AccessControlError::TransferInProgress);
     }
 
--- a/packages/access/src/ownable/storage.rs
+++ b/packages/access/src/ownable/storage.rs
@@ -5,7 +5,7 @@ use crate::{
         emit_ownership_renounced, emit_ownership_transfer, emit_ownership_transfer_completed,
         OwnableError,
     },
-    role_transfer::{accept_transfer, transfer_role},
+    role_transfer::{accept_transfer, transfer_role, PendingTransfer},
 };
 
 /// Storage keys for `Ownable` utility.
@@ -51,6 +51,9 @@ pub fn set_owner(e: &Env, owner: &Address) {
 }
 
 /// Initiates a 2-step ownership transfer to a new owner.
+/// Owner privileges for the current owner are not revoked until the
+/// recipient accepts the transfer.
+/// Overrides the previous pending transfer if there is one.
 ///
 /// # Arguments
 ///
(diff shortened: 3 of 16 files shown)


packages/contract-utils/src/upgradeable/mod.rs
```diff
@@ -23,52 +19,103 @@
...
+//!     #[only_role(operator, "manager")]
+//!     pub fn migrate(e: &Env, migration_data: Data, operator: Address) {
+//!         upgradeable::run_migration(e, || {
+//!             e.storage().instance().set(&symbol_short!("DATA_KEY"), &migration_data);
+//!         });
+//!     }
+//! }
+//! ```
 //!
+//! ## Migration Guide from `#[derive(Upgradeable)]`
```
ozgunozerk: I don't think this comment should live in here. Because it will be outdated. For example, say we are on `v2.X.X`, this doesn't make sense to keep. It may make sense to keep now, but the question becomes, when to remove this comment? 

These comments (migration for breaking changes) should live in the release notes imo. At least, that is how I navigated my breaking change fixes when I was using other libraries (like tokio), and I think it works well

ozgunozerk (changes requested):
Looks much much better. Love it!

I have 2 additional points:
- considering the community, they may not be familiar with what is `upgrade` and what is `migrate`. If my understanding is correct, `upgrade` is updating the functions of the contract, whilst `migrate` transfers the existing storage data from the old one to the new one. I think this will be very helpful to mention in the `mod.rs` and also in our docs. Because without this big picture, everything will be just confusing
- I really liked the new trait, but I'm honestly not a fan of implementing `migrate` function right into the contract. I'm curious about the reason/s why you prefer that way, instead of:
    - having the `migrate()` function in the same trait with `upgrade`
    - having the `migrate()` function in a separate trait

ozgunozerk:
Great work, I think this is becoming much better and streamlined 🚀 

Aside from 2 new nitpicks, all the new changes are looking great to me!

I still have the concern that I have mentioned in my previous concern though:

> I really liked the new trait, but I'm honestly not a fan of implementing migrate function right into the contract. I'm curious about the reason/s why you prefer that way, instead of:
having the migrate() function in the same trait with upgrade
having the migrate() function in a separate trait

ozgunozerk:
> Migration is most likely not happening through a single entrypoint. A single migrate function makes sense in the case of migrating only bounded data in one go. Furthermore, such an interface is forcing devs to "pack" all the arguments in a single #[contracttype] struct, while they usually need more flexibility in those circumstances. From this POV, adding a standard migrate function seems an overkill.

Good reasons. It may be worth to put this as comment in the code as well (mentioning it cannot be standardized, and the downsides you mention here).

PR #592, opened 2026-02-26
ZkEmail: DKIM registry

### LINKED ISSUE #557: ZkEmail: DKIM registry

brozorec:
Fixes #557 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added new zk-email package to workspace.
  * Introduced DKIM registry functionality enabling registration and revocation of DKIM public key hashes.
  * Added support for validating DKIM keys against registered hashes with automatic TTL management.
  * Implemented event emission for key registration and revocation activities.

* **Tests**
  * Comprehensive test suite validating DKIM registry operations including registration, validation, and revocation workflows.

### DIFF
--- a/packages/zk-email/src/dkim_registry/mod.rs
+++ b/packages/zk-email/src/dkim_registry/mod.rs
@@ -1,9 +1,41 @@
 //! # DKIM Registry Module
 //!
-//! This module provides the foundational trust anchor for the zkEmail
-//! ecosystem. It stores which DKIM public key hashes are valid for which email
-//! domains. All downstream modules (TBD) query the registry via
-//! [`is_key_hash_valid`].
+//! ## What is DKIM?
+//!
+//! **DKIM (DomainKeys Identified Mail)** is an email authentication standard.
+//! When a server sends an email, it signs the message with a **private RSA
+//! key**. The corresponding **public key** is published in the domain's DNS
+//! records. Email clients verify the signature against that public key to
+//! confirm the email genuinely came from that domain and wasn't tampered with.
+//!
+//! Example: when Gmail sends an email from `@google.com`, it attaches a DKIM
+//! signature. Mail clients fetch Google's public key from DNS and verify the
+//! signature is valid.
+//!
+//! ## What is zkEmail?
+//!
+//! **zkEmail** enables proving facts about an email without revealing the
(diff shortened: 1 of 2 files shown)


packages/zk-email/src/dkim_registry/mod.rs
```diff
@@ -0,0 +1,234 @@
+//! # DKIM Registry Module
```
ozgunozerk: This is a fairly new concept, and I think for this module, even less people compared to other concepts (acces-control, governance, etc.), will be familiarized with the concept of `DKIM` and how `zk` plays a role in this context.

```suggestion
//! # DKIM Registry Module
//! 
//! ## What is DKIM?
//! 
//! **DKIM (DomainKeys Identified Mail)** is an email authentication standard. When a server sends an email, it signs the message with a **private RSA key**. The corresponding **public key** is published in the domain's DNS records. Email clients verify the signature against that public key to confirm the email genuinely came from that domain and wasn't tampered with.
//! 
//! Example: when Gmail sends you an email from `@google.com`, it attaches a DKIM signature. Your mail client fetches Google's public key from DNS and verifies the signature is valid.
//! 
//! ## What is zkEmail?
//! 
//! **zkEmail** lets you prove facts about an email — _without revealing the email itself_ — using zero-knowledge proofs.
//! 
//! The core idea: since emails are DKIM-signed, you can construct a ZK proof that says _"I have a valid DKIM-signed email from `@google.com` that contains X"_ — provable on-chain without exposing the raw email content.
//! 
//! Use cases include:
//! 
//! - Prove you own a specific email address
//!     
//! - Prove you received a payment confirmation from a bank
//!     
//! - Prove an email from a DAO's domain authorized an action
//!     
//! 
//! ## What does this Module implement?
//! 
//! This is a **DKIM Registry** — the on-chain trust anchor for a zkEmail system. Its job is to _store which DKIM public key hashes are considered valid for which domains_.
//! 
//! When a ZK proof is verified on-chain, the verifier calls `is_key_hash_valid(domain_hash, public_key_hash)` to confirm the key used to sign the email is still trusted (not rotated or compromised).
```

So, we provide a starting point on the big picture, and then people can read the details below

ozgunozerk (changes requested):

PR #597, opened 2026-02-26
RWA: domain separators in claim issuer and metadata tests

### LINKED ISSUE #596: RWA: Domain-Level Hash Collision Between Claim Message and Claim Identifier

brozorec:
Fixes #596 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation to detect invalid metadata key and value types, preventing potential misconfiguration errors.

* **Documentation**
  * Updated message format documentation to clarify domain separation scheme and improve specification consistency.

### DIFF
--- a/packages/tokens/src/rwa/claim_issuer/test.rs
+++ b/packages/tokens/src/rwa/claim_issuer/test.rs
@@ -19,13 +19,15 @@ use soroban_sdk::{
 
 use crate::rwa::{
     claim_issuer::{
+        build_claim_identifier,
         storage::{
-            allow_key, decode_claim_data_expiration, encode_claim_data_expiration,
-            get_current_nonce_for, get_keys_for_topic, get_registries, invalidate_claim_signatures,
-            is_authorized_for, is_claim_expired, is_claim_revoked, is_key_allowed_for_registry,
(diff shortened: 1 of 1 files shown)


ozgunozerk (commented):
I see how domain level collision is fixed, but I couldn't see how country data validation is fixed? 

There are new tests indeed, yes. But I believe the audit issue was about validating country data before storing that information in storage.

Maybe I missed, but couldn't locate in the diff view any code related to that logic?

I know `validate_country_data` is present in other parts of the code (not in the diff view), but then what was the audit finding? Seeing the audit finding, but not being able to locate the fix in the diff view confused me.

brozorec:
> Maybe I missed, but couldn't locate in the diff view any code related to that logic?

Actually, it was fixed with https://github.com/OpenZeppelin/stellar-contracts/pull/511 when `metadata` was validated. Applying `.iter()` on the `Map` deserializes it and forces type-checking. I added just tests to make sure we don't miss those cases if we re-factor `validate_country_data()` in the future.

ozgunozerk:
Ok, I think we should tag the relevant pr then for that issue, not this one.

PR #605, opened 2026-03-05
Governor: fix re-exports

brozorec:
Following the library conventions, this PR re-exports all public functions from governor storage.

## Release Notes

* **Refactor**
  * Reorganized and expanded the governor module's public API to expose a comprehensive set of governance operations. Newly accessible functionality includes voting mechanisms, proposal lifecycle management, configuration management through getters and setters, and state querying capabilities. This restructuring improves accessibility and integration of core governance operations throughout the platform.

### DIFF
--- a/examples/fungible-governor/governor/src/governor.rs
+++ b/examples/fungible-governor/governor/src/governor.rs
@@ -1,8 +1,5 @@
 use soroban_sdk::{contract, contractimpl, Address, BytesN, Env, String, Symbol, Val, Vec};
-use stellar_governance::governor::{
-    storage::{self, set_token_contract},
-    Governor, ProposalState,
-};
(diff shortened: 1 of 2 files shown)


ozgunozerk (commented):
I'm not against the changes in `examples` folder. 

But I think we shouldn't touch the mod.rs, and fix the rest of the library. 

Due to, these are internal functions, so other crates importing them as `module::storage::{func1, func2, func3}` is not a disadvantage, and signals the message clearer imo. 

The downside of the suggested approach in this PR is, if we are exporting these, then why are we using `storage::func` in the trait name? It is confusing. 

We know to reason is to show that the trait implementation should not confuse the people, and clearly show that it is not recursive.

But it looks weird to export the function at crate level, then not using it directly, but with the `storage::func` path inside that very file

PR #650, opened 2026-03-23
RWA: country compliance modules

pasevin:
## Country Compliance Modules

This PR introduces two country-based compliance modules for RWA token management:

### Country Allow
Only recipients whose identity has at least one country code present in the allowlist may receive tokens. Administrators can configure which countries are allowed per token, either individually or in batch.

- T-REX reference: [CountryAllowModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/CountryAllowModule.sol)

### Country Restrict
Recipients whose identity has a country code on the restriction list are blocked from receiving tokens. Administrators can configure which countries are restricted per token, either individually or in batch.

- T-REX reference: [CountryRestrictModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/CountryRestrictModule.sol)


examples/rwa-country-allow/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+#[contract]
+pub struct CountryAllowContract;
+
+fn set_admin(e: &Env, admin: &Address) {
+    e.storage().instance().set(&DataKey::Admin, admin);
+}
+
+fn get_admin(e: &Env) -> Address {
+    e.storage().instance().get(&DataKey::Admin).expect("admin must be set")
+}
```
ozgunozerk: We usually implement them for the contract itself instead of free-functions, feel free to take a look at our other examples

examples/rwa-country-allow/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[contractimpl]
+impl CountryAllowContract {
+    pub fn __constructor(e: &Env, admin: Address) {
+        set_admin(e, &admin);
+    }
+}
+
+#[contractimpl(contracttrait)]
+impl CountryAllow for CountryAllowContract {
```
ozgunozerk: I think all the implementations below can be skipped with default implementations. The only problem may be `require_module_admin_or_compliance_auth`, but I think we can workaround that. In fact, that limitation itself requires a bit discussion imo

examples/rwa-country-allow/README.md
```diff
@@ -0,0 +1,50 @@
+# Country Allow Module
```
ozgunozerk: I don't see the necessity of this readme which is specific to `rwa-country-allow`. We have / will have enough inline docs in the module itself. We are not putting a readme in examples folder, unless it requires additional documentation due to combining things and explaining the workflow for contract deployment / communication (as in upgradeable)

ozgunozerk (commented):
examples are not done in the light of our other examples. The structure is missing, there is only lib file, and an unwanted readme. 
There isn't contract, nor tests.

pasevin:
> examples are not done in the light of our other examples. The structure is missing, there is only lib file, and an unwanted readme. There isn't contract, nor tests.

Thanks, I updated this to be closer to the structure of the other examples.

The example now has the usual split, the extra README is gone, and I added local tests for both country examples so they’re easier to review and a bit more complete.

I left the repeated methods as-is for now instead of trying to refactor that part in this PR. That felt like a bigger cleanup than I wanted to mix into what was meant to stay a structural-only pass. If you want, I can take that on next in a separate follow-up.

examples/rwa-country-allow/src/lib.rs
ozgunozerk: This is still not addressed

ozgunozerk:
Why there is no trait on `mod.rs` that can be implemented by the contract, why instead we are implementing everything on the contract itself? 

I think this doesn't match with the design of the library.

examples/rwa-country-allow/src/contract.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    fn on_destroyed(_e: &Env, _from: Address, _amount: i128, _token: Address) {}
+
+    fn can_transfer(e: &Env, _from: Address, to: Address, _amount: i128, token: Address) -> bool {
+        country_allow::can_transfer(e, &to, &token)
+    }
+
+    fn can_create(e: &Env, to: Address, _amount: i128, token: Address) -> bool {
+        country_allow::can_transfer(e, &to, &token)
+    }
```
ozgunozerk: why use `can_transfer` in both? These two functions are definitely serving different purposes. If you think it is ok to use the internal `country_allow::can_transfer` for both, please provide good reasons on why it is safe, and also rename it so it wouldn't be confusing

ozgunozerk (changes requested):

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,195 @@
...
+};
+
+#[contracttype]
+#[derive(Clone)]
+pub enum CountryAllowStorageKey {
+    /// Per-(token, country) allowlist membership entry.
+    AllowedCountry(Address, u32),
+}
+
+// ################## RAW STORAGE ##################
```
ozgunozerk: we don't have sections such as raw storage, actions, etc. 

I thought the code quality skill would have caught them, but it's ok. You can take a look at other `storage.rs` examples or take a look at the code quality skill itself, the sections should be present there

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,195 @@
...
+/// * `token` - The token address.
+/// * `country` - The ISO 3166-1 numeric country code to allow.
+///
+/// # Security Warning
+///
+/// This helper performs no authorization checks.
+pub fn set_country_allowed(e: &Env, token: &Address, country: u32) {
+    let key = CountryAllowStorageKey::AllowedCountry(token.clone(), country);
+    e.storage().persistent().set(&key, &());
+    e.storage().persistent().extend_ttl(&key, MODULE_TTL_THRESHOLD, MODULE_EXTEND_AMOUNT);
```
ozgunozerk: we don't extend ttl after `set`, did code quality skill not catch it?

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,195 @@
...
+    _amount: i128,
+    token: &Address,
+) -> bool {
+    can_receive(e, to, token)
+}
+
+/// Returns `true` if the mint recipient has at least one allowed country.
+pub fn can_create(e: &Env, to: &Address, _amount: i128, token: &Address) -> bool {
+    can_receive(e, to, token)
+}
```
ozgunozerk: why the inline doc format is not followed for these 2 functions?

packages/tokens/src/rwa/compliance/modules/country_allow/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `to` - The recipient address.
+    /// * `amount` - The minted amount.
+    /// * `token` - The token address.
+    fn can_create(e: &Env, to: Address, amount: i128, token: Address) -> bool {
+        storage::can_create(e, &to, amount, &token)
+    }
```
ozgunozerk: aren't these exposed from `pub trait ComplianceModule` already? 

I think it may make more sense to modify this trait (and all the other modules you are writing) in this way:

```rust
pub trait CountryAllow: ComplianceModule
```

so that, all the compliance hook methods will be there. I think this is a safer approach so that no hooks will be forgotten, and also better for eliminating duplication. This comment is probably the most important one for this PR, because it includes design choice

ozgunozerk (changes requested):
Much better! 

I found 2-3 small things, that I expected code-quality skill to catch tbh. If it missed them, can you please let me know, so I can tweak the skill further for emphasizing these points? 

Thanks!

pasevin:
> Much better!
> 
> I found 2-3 small things, that I expected code-quality skill to catch tbh. If it missed them, can you please let me know, so I can tweak the skill further for emphasizing these points?
> 
> Thanks!

Thanks again for the careful review. I addressed the latest comments and aligned the country modules more closely with the code-quality rules and nearby library patterns.

Specifically, I updated the storage sections/docs, removed TTL extension after writes, routed events through module-level emit helpers, and changed the country traits to compose with `ComplianceModule` so the shared hook surface isn’t duplicated.

I hope this is up to par now :)

ozgunozerk (approved):
LGTM! 
I'll review with a fresh set of eyes tomorrow morning just to double check everything. 

In the meantime, there is only one small issue remaining:
- in mod.rs files, the trait should come first, then the event related things. It is just re-ordering

Feel free to adjust other PRs based on the feedback for this PR.

And, in the meantime, if you think the code-quality skill missed some important corrections that I've made in here, please let me know, I'll tweak it :)

Thanks again!

pasevin:
> LGTM! I'll review with a fresh set of eyes tomorrow morning just to double check everything.
> 
> In the meantime, there is only one small issue remaining:
> 
> * in mod.rs files, the trait should come first, then the event related things. It is just re-ordering
> 
> Feel free to adjust other PRs based on the feedback for this PR.
> 
> And, in the meantime, if you think the code-quality skill missed some important corrections that I've made in here, please let me know, I'll tweak it :)
> 
> Thanks again!

Thanks for your patience, it has been a great learning curve!

So far I would say the skill was a great addition. Maybe the most useful addition from my learning would be:

```
If a PR introduces a new module pattern, compare it against at least two nearby native examples and flag any differences in trait composition, event routing, storage sections, TTL policy, auth shape, and doc structure.
```

But it's up to you!

examples/rwa-country-allow/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+#[contract]
+pub struct CountryAllowContract;
+
+fn set_admin(e: &Env, admin: &Address) {
+    e.storage().instance().set(&DataKey::Admin, admin);
+}
+
+fn get_admin(e: &Env) -> Address {
+    e.storage().instance().get(&DataKey::Admin).expect("admin must be set")
+}
```
ozgunozerk: `access_control` module exposes `set_admin` and `get_admin`, let's use those instead

packages/tokens/src/rwa/compliance/modules/country_allow/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// * `e` - Access to the Soroban environment.
+    /// * `token` - The token whose IRS is being configured.
+    /// * `irs` - The Identity Registry Storage contract address.
+    ///
+    /// # Notes
+    ///
+    /// No default implementation is provided because this is a privileged
+    /// operation that requires custom access control. Access control should be
+    /// enforced before calling
+    /// [`super::storage::set_irs_address`].
```
ozgunozerk: `super` usage is frowned upon in this library. Let's use the full path

ozgunozerk: also, let's search for `super` keyword for this PR and change all the instances of it with the full path

packages/tokens/src/rwa/compliance/modules/country_allow/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+use super::ComplianceModule;
+
+/// Country allowlist compliance module trait.
+///
+/// This trait defines the contract-facing API for the country allowlist module.
+/// Low-level state changes live in [`storage`]. Privileged methods have no
+/// default implementation because each contract must enforce its own access
+/// control before delegating to storage helpers.
+#[contracttrait]
+pub trait CountryAllow: ComplianceModule {
```
ozgunozerk, line 24: the trait methods inline documentation is not matching with this library's style:
- error section is missing
- event section style is not matching

can you fix these for all the trait methods in this PR?

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,221 @@
...
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `account` - The account whose country data is checked.
+/// * `token` - The token address.
+///
+/// # Errors
+///
+/// * [`crate::rwa::compliance::modules::ComplianceModuleError::IdentityRegistryNotSet`]
```
ozgunozerk: `ComplianceModuleError` should have been imported above. No need to provide the superfluous path here, it is way too long. Again, apply this comment to everywhere possible please

ozgunozerk (changes requested):
some styling changes are required, code itself looks good

pasevin:
> some styling changes are required, code itself looks good

Thanks, I've applied the changes

packages/tokens/src/rwa/compliance/modules/country_allow/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `token` - The token whose IRS is being configured.
+    /// * `irs` - The Identity Registry Storage contract address.
+    ///
+    /// # Errors
+    ///
+    /// Implementations should fail if the caller is not authorized to configure
+    /// the IRS address.
```
ozgunozerk: There is no error specified in here? I'm confused, why is there such an explanation? This doesn't exist in anywhere in our library?

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+/// # Errors
+///
+/// * [`ComplianceModuleError::IdentityRegistryNotSet`] - When no IRS has been
+///   configured for `token`.
+///
+/// # Cross-Contract Calls
+///
+/// Calls the IRS to resolve country data for `account`.
+///
+/// [`ComplianceModuleError::IdentityRegistryNotSet`]: crate::rwa::compliance::modules::ComplianceModuleError::IdentityRegistryNotSet
```
ozgunozerk: I mentioned in my previous comments that the full path is superfluous. Now, the link itself is fixed, but the full path is put into the explanation part instead. What is the reason for this?

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+        set_country_allowed(e, token, country);
+        emit_country_allowed(e, token, country);
+    }
+}
+
+/// Removes a country from the allowlist for `token`.
+///
+/// Deletes the membership entry and emits
+/// [`crate::rwa::compliance::modules::country_allow::CountryUnallowed`] if the
+/// country was currently allowed.
```
ozgunozerk: Why there is no `Events` section for storage functions as well?

ozgunozerk (changes requested):

packages/tokens/src/rwa/compliance/modules/country_allow/storage.rs
ozgunozerk: Also, this is an error. It should have been in the `# Errors` section. Why has it moved to `# Cross-Contract Calls` section?

pasevin: Thanks, I see the issue now.

What tripped me up was that `ComplianceModuleError` is only raised indirectly here through `get_irs_country_data_entries`, not directly in these functions. I first tried importing the error to keep the rustdoc link short, but since the type was only used in docs, clippy flagged it as an unused import. Then I tried a reference-style rustdoc link, but that still left the full path in the docs, just lower down.

Looking at sibling modules, the pattern seems to be: import the error type only when the file uses it in code, and otherwise document delegated errors as `refer to [...] errors`.

So I think the cleaner fix here is to write:

`refer to [get_irs_country_data_entries] errors`

instead of linking `ComplianceModuleError::IdentityRegistryNotSet` directly from these wrappers.

Does that match what you’d expect?

PR #651, opened 2026-03-23
RWA: balance and supply compliance modules

pasevin:
## Balance & Supply Compliance Modules

This PR introduces two balance/supply-related compliance modules for RWA token management:

### Max Balance
Tracks effective balances per identity (not per wallet), enforcing a per-token cap. When a transfer or mint would push an identity's aggregate balance above the configured maximum, the operation is blocked. Balances can be pre-seeded for migration scenarios.

- T-REX reference: [MaxBalanceModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/MaxBalanceModule.sol)

### Supply Limit
Enforces a global supply cap per token. The module maintains an internal supply counter that increments on mint and decrements on burn. When a mint would push the total above the configured limit, the operation is blocked.

- T-REX reference: [SupplyLimitModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/SupplyLimitModule.sol)


brozorec (changes requested):
Actually, we don't need an additional trait per module: we already have `ComplianceModule` that must be implemented by each.

In the current state, some of the logic is contained in the default functions from `mod.rs` and some in `storage.rs`. Our philosophy is to provide public storage functions to be composed by implementors, and only wherever it makes sense to provide defaults. Here, every module has its own logic so defaults won't work.

 Following lib's general pattern, I'd rather suggest to:
- move all the specific implementations from the traits to `storage.rs` as "free" functions
- keep `mod.rs` for just events, constants and re-export
- in "examples", every module implements `trait ComplianceModule` and comes with an additional `contractimpl` for its specific functions (no need to implement a trait), and use there the free functions from `storage.rs`.

pasevin:
> Actually, we don't need an additional trait per module: we already have `ComplianceModule` that must be implemented by each.
> 
> In the current state, some of the logic is contained in the default functions from `mod.rs` and some in `storage.rs`. Our philosophy is to provide public storage functions to be composed by implementors, and only wherever it makes sense to provide defaults. Here, every module has its own logic so defaults won't work.
> 
> Following lib's general pattern, I'd rather suggest to:
> 
> * move all the specific implementations from the traits to `storage.rs` as "free" functions
> * keep `mod.rs` for just events, constants and re-export
> * in "examples", every module implements `trait ComplianceModule` and comes with an additional `contractimpl` for its specific functions (no need to implement a trait), and use there the free functions from `storage.rs`.

Thanks @brozorec, I've aligned the PR with the library conventions and my understanding.

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `token` - The token address.
+/// * `value` - The maximum balance per identity.
+pub fn set_max_balance(e: &Env, token: &Address, value: i128) {
+    let key = MaxBalanceStorageKey::MaxBalance(token.clone());
+    e.storage().persistent().set(&key, &value);
+    e.storage().persistent().extend_ttl(&key, MODULE_TTL_THRESHOLD, MODULE_EXTEND_AMOUNT);
```
brozorec: no need to extend ttl when setting an entry

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    if from_id == to_id {
+        return;
+    }
+
+    let from_balance = get_id_balance(e, token, &from_id);
+    assert!(
+        can_increase_identity_balance(e, token, &to_id, amount),
+        "MaxBalanceModule: recipient identity balance exceeds max"
+    );
```
brozorec: Use `panic_with_error` instead `assert`. Errors are to be defined in mod.rs and enumerated without overlaps with other modules.

packages/tokens/src/rwa/compliance/modules/max_balance/mod.rs
```diff
@@ -0,0 +1,32 @@
+//! Max balance compliance module — Stellar port of T-REX
+//! [`MaxBalanceModule.sol`][trex-src].
+//!
+//! Tracks effective balances per **identity** (not per wallet), enforcing a
+//! per-token cap.
+//!
+//! [trex-src]: https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/MaxBalanceModule.sol
```
brozorec, line 7: The T-REX link is fine for a reference, but we might want to provide more context and provide a description of the module here as well.

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+/// * `token` - The token address.
+/// * `identity` - The on-chain identity address.
+/// * `balance` - The new balance value.
+pub fn set_id_balance(e: &Env, token: &Address, identity: &Address, balance: i128) {
+    let key = MaxBalanceStorageKey::IDBalance(token.clone(), identity.clone());
+    e.storage().persistent().set(&key, &balance);
+    e.storage().persistent().extend_ttl(&key, MODULE_TTL_THRESHOLD, MODULE_EXTEND_AMOUNT);
+}
+
+// ################## HELPERS ##################
```
brozorec: helpers usually go at the end

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    let current = get_id_balance(e, token, identity);
+    add_i128_or_panic(e, current, amount) <= max
+}
+
+// ################## ACTIONS ##################
+
+/// Validates, stores, and emits [`MaxBalanceSet`] for the given cap.
+///
```
brozorec: It might be worth documenting that when max balance is not set (ie `get_max_balance` returns 0), it's bound to token's total supply (this is how I understand L102 above).

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `token` - The token address.
+/// * `max` - The maximum balance per identity.
```
brozorec, line 106: Add SECURITY WARNING section that this fn has to be invoked after authorization and access control check.

brozorec: This should be added to all state-modifying functions that delegate auth checks to the implementor.

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+        return true;
+    }
+
+    let current = get_id_balance(e, token, identity);
+    add_i128_or_panic(e, current, amount) <= max
+}
+
+// ################## ACTIONS ##################
+
+/// Validates, stores, and emits [`MaxBalanceSet`] for the given cap.
```
brozorec: Events must be documented in a separate Events docstring sections

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -0,0 +1,315 @@
...
+    identities: &Vec<Address>,
+    balances: &Vec<i128>,
+) {
+    assert!(
+        identities.len() == balances.len(),
+        "MaxBalanceModule: identities and balances length mismatch"
+    );
+    for i in 0..identities.len() {
+        let id = identities.get(i).unwrap();
+        let bal = balances.get(i).unwrap();
```
brozorec: `.get_unchecked()` comes more handy in such cases

packages/tokens/src/rwa/compliance/modules/supply_limit/storage.rs
```diff
@@ -0,0 +1,198 @@
...
+};
+
+#[contracttype]
+#[derive(Clone)]
+pub enum SupplyLimitStorageKey {
+    /// Per-token supply cap.
+    SupplyLimit(Address),
+    /// Per-token internal supply counter (updated via hooks).
+    InternalSupply(Address),
+}
```
brozorec, line 22: It might be useful to document the difference/relation between both.

packages/tokens/src/rwa/compliance/modules/supply_limit/storage.rs
```diff
@@ -0,0 +1,198 @@
...
+/// * `token` - The token address.
+pub fn get_supply_limit(e: &Env, token: &Address) -> i128 {
+    let key = SupplyLimitStorageKey::SupplyLimit(token.clone());
+    e.storage()
+        .persistent()
+        .get(&key)
+        .inspect(|_: &i128| {
+            e.storage().persistent().extend_ttl(&key, MODULE_TTL_THRESHOLD, MODULE_EXTEND_AMOUNT);
+        })
+        .unwrap_or_default()
```
brozorec, line 40: Does returning a default value make sense instead of panicking when not set? The whole module is about this single value, so if an implementor binds to this compliance module, they should configure it properly imo.

examples/rwa-max-balance/src/contract.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[contractimpl]
+impl MaxBalanceContract {
+    pub fn __constructor(e: &Env, admin: Address) {
+        set_admin(e, &admin);
+    }
+
+    pub fn set_identity_registry_storage(e: &Env, token: Address, irs: Address) {
+        require_module_admin_or_compliance_auth(e);
```
brozorec: I can't think of cases in which authorization for setters comes from compliance 🤔 Here, we might assume the authorization is from admin, while in hooks it's from the compliance contract. As a ref, the T-REX implementation checks compliance's owner in setters.

brozorec (changes requested):
Hey @pasevin overall logic of the modules seems fine, but there are some points left to be addressed about the structure, docstring composition and the error handling.

pasevin:
Thanks for the review @brozorec , I addressed the style/structure items from the comments and based on the learnings from #650 

Mainly:
- removed TTL extension after writes
- aligned storage sections and helper placement
- added event helpers and proper `# Events` docs
- added security warnings for unauthenticated storage helpers
- replaced the relevant `assert!` calls with explicit contract errors
- cleaned up the example admin helpers and related tests

I left the remaining threads unresolved on purpose because they’re more about behavior/auth semantics than style. I’ll look at those next separately so we don’t mix design decisions into the styling pass.

packages/tokens/src/rwa/compliance/modules/supply_limit/storage.rs
pasevin: I aligned this with T-REX semantics: an unset limit effectively behaves as `0`, so positive mints fail closed rather than treating unset as “unlimited”. I kept `get_supply_limit()` returning `0` as the raw query helper, and added `get_supply_limit_or_panic()` for callers that need to distinguish missing configuration. Do you think we should intentionally diverge from T-REX here and make the validation path panic when the limit is not configured?

examples/rwa-max-balance/src/contract.rs
pasevin: Addressed in the latest version: configuration setters are now admin-gated, while state-changing hooks require authorization from the configured compliance contract. This actually was a miss on my end, thanks for catching!

I kept the trait API without an explicit `operator` for now, since adding that would be a broader public API shape decision; happy to adjust if you prefer the same explicit-operator pattern used by other modules.

packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
pasevin: This changed in the latest version: unset max balance no longer means “bound to total supply” or “uncapped”. To follow T-REX’s default mapping behavior, the raw value is 0, and positive balance increases fail closed unless a max balance is configured. I’ll make sure the docs state that clearly so get_max_balance() == 0 is not read as an unlimited/default-to-total-supply case.

pasevin: This changed in the latest version: unset max balance no longer means “bound to total supply” or “uncapped”. To follow T-REX’s default mapping behavior, the raw value is `0`, and positive balance increases fail closed unless a max balance is configured. I’ve clarified the docs so `get_max_balance() == 0` is not read as an unlimited/default-to-total-supply case.

PR #652, opened 2026-03-23
RWA: transfer compliance modules

pasevin:
### Transfer Restrict
Maintains a per-token address allowlist. Transfers are permitted when the sender is allowlisted; otherwise the recipient must be allowlisted.

- T-REX reference: [TransferRestrictModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/TransferRestrictModule.sol)

### Initial Lockup Period
Enforces a lockup window on tokens received through primary issuance. The module tracks per-wallet balances and lock entries, and transfers are allowed only when the sender has sufficient unlocked balance.

- T-REX reference: [TimeExchangeLimitsModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/TimeExchangeLimitsModule.sol)

### Time Transfers Limits
Enforces cumulative transfer limits within configurable time windows. The module tracks each identity's transfer volume inside each active window and rejects transfers that would exceed any configured limit.

- T-REX reference: [TimeTransfersLimitsModule.sol](https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/TimeTransfersLimitsModule.sol)


pasevin:
Captured a few follow-up/design notes from the broader style pass:

1. Error-code numbering now leaves room for the shared variants already used in adjacent RWA module branches. This is intentional for stack consistency, but if this PR is reviewed as fully standalone we can renumber the new variants contiguously.
2. The new module-specific traits follow the current RWA pattern and omit an explicit `operator: Address`. Broader non-RWA traits often include `operator`, so this remains a design point if maintainers prefer the explicit-operator API shape here too.
3. The standalone example crates keep the existing RWA standalone shape (`rlib` plus public `contract` module) so generated clients/tests work consistently with the other RWA standalone examples. We can tighten that if maintainers want these to follow the narrower non-RWA example crate shape.

PR #683, opened 2026-04-08
Governance: timelock trait

brozorec:
## Summary by CodeRabbit

* **New Features**
  * Added a public Timelock trait interface for timelock controllers and support for scheduling/executing/cancelling operations via that trait.

* **Bug Fixes**
  * Governor now records timelock operation IDs and will cancel corresponding timelock operations when proposals are cancelled.

* **Refactor**
  * Timelock controller methods moved to trait-based implementations and several internal query helpers were consolidated.

* **Tests**
  * Test setup updated to grant the governor additional timelock permissions for cancellation flows.


packages/governance/src/timelock/mod.rs
```diff
@@ -59,6 +67,259 @@ pub use crate::timelock::storage::{
...
+
+    /// Returns the current state of an operation.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `operation_id` - The unique identifier of the operation.
+    fn get_operation_state(e: &Env, operation_id: BytesN<32>) -> OperationState {
+        storage::get_operation_state(e, &operation_id)
+    }
```
ozgunozerk, line 128: Since we have this one, do we need `operation_exists/pending/done` etc. as public-facing methods? I see the value of having these in storage, because for example, we check if the operation is pending before cancelling it, but I'm not sure how much value do they bring whilst making the trait more convoluted. 

I'm not saying we should get rid of this, but want to discuss to shed some more light on this

packages/governance/src/timelock/mod.rs
```diff
@@ -59,6 +67,259 @@ pub use crate::timelock::storage::{
...
+    /// * topics - `["operation_scheduled", id: BytesN<32>, target: Address]`
+    /// * data - `[function: Symbol, args: Vec<Val>, predecessor: BytesN<32>,
+    ///   salt: BytesN<32>, delay: u32]`
+    ///
+    /// # Notes
+    ///
+    /// * Authorization for `proposer` is required.
+    /// * The implementer must verify that `proposer` has the appropriate role,
+    ///   construct an [`Operation`], and call [`schedule_operation`].
+    #[allow(clippy::too_many_arguments)]
```
ozgunozerk, line 186: I know the example code for implementation of this is given in the above, but i see more value in having the example/suggested implementation right here in the inline docs of the function itself. I'd suggest moving it here rather than duplicating it. The same goes for `execute_op`

ozgunozerk (commented):
One issue I missed previously. 

Why are we emitting `executed` event inside `set_execute_operation` whilst this one does not in fact executes it. Shouldn't we emit the event in here as the last line after `invoke_contract` call?

```
pub fn execute_operation(e: &Env, operation: &Operation) -> Val {
    set_execute_operation(e, operation);

    e.invoke_contract::<Val>(&operation.target, &operation.function, operation.args.clone())
}
```

packages/governance/src/timelock/mod.rs
```diff
@@ -59,6 +67,259 @@ pub use crate::timelock::storage::{
...
+    /// * data - `[function: Symbol, args: Vec<Val>, predecessor: BytesN<32>,
+    ///   salt: BytesN<32>]`
+    ///
+    /// # Notes
+    ///
+    /// * Authorization for `executor` is optional (open execution is allowed
+    ///   when `executor` is `None`).
+    /// * The implementer must construct an [`Operation`] and call
+    ///   [`execute_operation`].
+    fn execute_op(
```
ozgunozerk: I'd suggest renaming it as `call_execute` or `invoke_execute`, because this itself, is not executing, it is just triggering/invoking the execute. 

This suggestion may sound like an overkill/nitpick, but I think clarifications in terms of roles and actions are quite important especially in such an involved module like governance

examples/fungible-governor-timelock/governor/src/governor.rs
```diff
@@ -90,28 +86,28 @@ impl Governor for GovernorTimelockContract {
...
+        let op_id = timelock_client.schedule_op(
+            &e.current_contract_address(),
+            &Symbol::new(e, "execute"),
+            &execute_args,
+            &zero_predecessor,
+            &description_hash,
+            &delay,
+            &e.current_contract_address(),
         );
+        e.storage()
```
ozgunozerk, line 108: manipulating the storage in function implementation felt like too low-level. 

Any other solution that we can put these calls into the storage functions? 

Was there a specific case for this example that requires storage manipulation to happen at this level?

packages/governance/src/timelock/mod.rs
brozorec: It mirrors the Solidity [implementation](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/9cfdccd35350f7bcc585cf2ede08cd04e7f0ec10/contracts/governance/TimelockController.sol#L175), but you make a valid point: it feels superfluous 🤔

packages/governance/src/timelock/mod.rs
brozorec: Do you mean these examples https://github.com/OpenZeppelin/stellar-contracts/blob/80c2d9200e71d28bbcce759c562ec278ea5edc66/packages/governance/src/timelock/mod.rs#L37

ozgunozerk: yes 👍

packages/governance/src/timelock/mod.rs
brozorec: > is not executing, it is just triggering/invoking the execute

Sorry, don't get the difference 😅 if it calls/invokes execute, it does execute it, ie. the desired effects are produced.

However, I agree that renaming them to just `schedule`, `execute`, `cancel` seems reasonable.

examples/fungible-governor-timelock/governor/src/governor.rs
brozorec: > manipulating the storage in function implementation felt like too low-level.

This manipulates its own storage, so I don't see a problem with this. 

> Any other solution that we can put these calls into the storage functions?

This mapping is needed on the conjunction between governor and timelock as in this example. I'm not sure it should be handled in either of the individual modules.

packages/governance/src/timelock/mod.rs
ozgunozerk: Ah sorry, I thought this was calling the `execute` from governor. But this in fact calls the target contract directly, and does not call an intermediate `execute` like `invoke_contract::execute`, 

But, I agree with `*_op` is unnecessary and we can rename the just to `schedule`, `execute`, and so on... 👍

examples/fungible-governor-timelock/governor/src/governor.rs
ozgunozerk: Just thinking from a very straightforward point, I'd expect the operation ID to be stored in timelock storage, and got managed there (in timelock's storage functions). 

Having the storage exposed in the contract feels like a bad design to me, especially for something like `OperationID`. 

I'd even suggest having the timelock contract key managed with a Timelock Extension Trait for Governor. Defining and managing storage in example contracts feels off to me. 

We can have a small huddle about this if you like to explore other options or evaluate this again if you think this is fine to leave as is, I'm definitely up for challenging my own view as well 👍

packages/governance/src/timelock/mod.rs
ozgunozerk: after giving it more thought, I'm more inclined to remove them, I don't think they are super useful to expose. If any dapp requires them, it would be very easy to write their custom logic by comparing the state imo. 

Don't have a strong opinion, you can make the final decision, I'm ok with both approaches

brozorec:
> One issue I missed previously.
> 
> Why are we emitting `executed` event inside `set_execute_operation` whilst this one does not in fact executes it. Shouldn't we emit the event in here as the last line after `invoke_contract` call?

The idea is that it can be called directly for self-administration scenarios where the timelock contract needs to execute operations on itself. `execute_operation` can't be called from within the timelock to execute a method on itself. If we move it in `execute_operation` we won't have an event in such cases.

examples/fungible-governor-timelock/governor/src/governor.rs
brozorec: I'm not shocked about a contract dealing with storage that regards its inner workings. The example contract is on the cross-road between timelock and governor so it needs to map data between those independent modules. That being said, I agree that an extension doing this job sounds as a viable option to explore and we should evaluate the eventual complexity it might introduced.

ozgunozerk: Ok, it is not a blocker so I'll resolve this convo but open an issue about this potential improvement. We can continue to discuss it there, and if we decide to implement it, it can be in another PR, all good!

ozgunozerk (approved):

PR #693, opened 2026-05-08
Power function for float raised by float

### LINKED ISSUE #692: 🏗️ [Core Feature]: Float raised by float

Implement a power function variant for float raised by float to our math library

ozgunozerk:
Fixes #692 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added natural logarithm (ln) and exponential (exp) mathematical functions
  * Introduced fractional power exponentiation (powf) to raise values to non-integer exponents
  * Integer power operations renamed from pow to powi for naming clarity
  * Added checked variants for safe error handling of ln, exp, and powf operations

* **Error Handling**
  * Improved validation for mathematical operations with invalid inputs

### DIFF
--- a/packages/contract-utils/src/math/exp_ln.rs
+++ b/packages/contract-utils/src/math/exp_ln.rs
@@ -24,6 +24,20 @@
 
 use soroban_sdk::{Env, I256, U256};
 
+/// Input threshold at/below which `exp_wad(x)` returns 0.
+///
+/// For `x` this small, `e^x * 10^18` evaluates to less than `0.5`, which
+/// truncates to `0` in WAD precision. The exact value is
+/// `floor(ln(0.5 * 10^-18) * 10^18)` ≈ `-42.139 * 10^18`.
+const EXP_INPUT_MIN: i128 = -42_139_678_854_452_767_551;
+
+/// Input threshold at/above which `exp_wad(x)` overflows.
+///
+/// For `x` this large, `e^x` exceeds the range Solmate's algorithm can
+/// represent in its internal i256-with-2^96-scale basis. The exact value
+/// is `floor(ln((2^255 - 1) / 10^18) * 10^18)` ≈ `135.305 * 10^18`.
+const EXP_INPUT_MAX: i128 = 135_305_999_368_893_231_589;
+
 /// Natural logarithm of a WAD value.
 ///
 /// Returns `None` if `x <= 0` or the result doesn't fit in `i128`.
@@ -68,7 +82,7 @@ pub(super) fn ln_wad(e: &Env, x: i128) -> Option<i128> {
     p = p.mul(&x).shr(96).sub(&I256::from_parts(e, 0, 0, 0xb9a025d814, 0xb29c212b8b1a07ce));
     // p = p * x - (795164235651350426258249787498 << 96)
     // (Pre-shifted constant; leaves p in 2^192 basis to skip a final shift.)
-    p = p.mul(&x).sub(&I256::from_parts(e, 10, 0x09507084cc699bb0, 0xe71ea86a00000000, 0));
+    p = p.mul(&x).sub(&I256::from_parts(e, 0xa, 0x09507084cc699bb0, 0xe71ea86a00000000, 0));
 
     // q polynomial — q has no zeros in the domain (all roots complex).
 
@@ -105,7 +119,7 @@ pub(super) fn ln_wad(e: &Env, x: i128) -> Option<i128> {
     r = r.add(
         &I256::from_parts(
             e,
-            2_644_146_654_357,
+            0x267a36c0c95,
             0xb3975ab3ee5b203a,
             0x7614a3f75373f047,
             0xd803ae7b6687f2b3,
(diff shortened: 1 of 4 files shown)


packages/contract-utils/src/math/exp_ln.rs
```diff
@@ -0,0 +1,225 @@
...
+/// matching Solmate). Returns `None` for `x >= 135305999368893231589`
+/// (would overflow even in i256-with-2^96-scale).
+pub(super) fn exp_wad(e: &Env, x: i128) -> Option<i128> {
+    // dec: -42139678854452767551 — below this, exp(x) rounds to 0 in WAD.
+    if x <= -42_139_678_854_452_767_551 {
+        return Some(0);
+    }
+    // dec: 135305999368893231589 — at/above this, the result exceeds i256
+    // headroom in Solmate's algorithm.
+    if x >= 135_305_999_368_893_231_589 {
```
brozorec: it might be a good idea to move up those values as constants and to explain what they represent (could be inferred from the const names or with a inline comment)

packages/contract-utils/src/math/exp_ln.rs
brozorec, line 1: - Is it intentional the mixup of decimal and hex represantions?
- What's the reasoning for both fn visibility, why not just `pub`? It might be useful for clients to access those lower-level functions.
- `# Arguments` sections are missing from both functions docs

packages/contract-utils/src/math/wad.rs
```diff
@@ -482,6 +488,207 @@ impl Wad {
...
+    ///
+    /// # Errors
+    ///
+    /// * [`SorobanFixedPointError::InvalidBase`] - When `self <= 0`. `ln` is
+    ///   undefined for non-positive inputs.
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// let e = Env::default();
```
brozorec:
```suggestion

```

That's a testing artifact, in production it's implicit so I suggest to remove it from this example and the others below.

brozorec (changes requested):
Everything looks sound, just couple of nitpicks. Would it be possible to further improve the coverage?

packages/contract-utils/src/math/exp_ln.rs
ozgunozerk: > Is it intentional the mixup of decimal and hex represantions?

It was due to: I256::from_parts(e, hi_hi: i64, hi_lo: u64, lo_hi: u64, lo_lo: u64) -> first one being `i64`, and the rest 3 being `u64`, but I agree it seems confusing. I'll think about it and change them all to hex or decimal. Thanks for pointing it out 👍 

> What's the reasoning for both fn visibility, why not just pub? It might be useful for clients to access those lower-level functions.

they are purely helper functions. We have the same structure in `i128_fixed_point.rs` and `i256_fixed_point.rs`. I don't see much value in having `ln` and `e^self` exposed as public functions, because the only usage of them in smart contract/finance setting to my knowledge, will be taking powers of floating numbers, for which we already exposed these as public. 

The main reason I made them private (or let's say crate-level public only), is due to `exp_wad`. People may get confused with `exp_wad` being available, and they may think this is the function they need for taking exponentiation. 

Since I didn't see much benefit of making them public, I kept them only crate-level public so that the `wad` file can use them.

> Arguments sections are missing from both functions docs

We don't have extensive inline docs for the helper functions, see `i128_fixed_point.rs`, the last 2 functions for example (actually the whole # HELPERS section):

```rust
/// Performs checked ceil(r / z)
fn checked_div_ceil(r: i128, z: i128) -> Option<i128> {
    if (r <= 0 && z > 0) || (r >= 0 && z < 0) {
        // ceiling is taken by default for a negative or zero result
        r.checked_div(z)
    } else {
        // floor taken by default for a positive result
        let remainder = r.checked_rem_euclid(z)?;

        // check for div overflow (i128::MIN / -1)
        r.checked_div(z)?.checked_add(if remainder > 0 { 1 } else { 0 })
    }
}

/// Performs ceil(r / z)
fn div_ceil(r: i128, z: i128) -> i128 {
    if (r <= 0 && z > 0) || (r >= 0 && z < 0) {
        // ceiling is taken by default for a negative or zero result
        r / z
    } else {
        // floor taken by default for a positive result
        let remainder = r.rem_euclid(z);

        r / z + (if remainder > 0 { 1 } else { 0 })
    }
}
```

packages/contract-utils/src/math/wad.rs
ozgunozerk: I think there is value in making it explicit. So I was thinking fixing the other instances of `# Examples` in this file w.r.t this. I searched the whole repo with `# Examples` and they were explicit enough pertaining to this.

brozorec: In production, you get it implicitly from the fn args and providing an example that defines it explicitly creates more confusion imo? What value are you seeing in making it explicit?

ozgunozerk: i was just thinking in the scope of the example code piece. We have chosen to name `Env` as `e` in our library, and I thought maybe some other people will not be that accustomed to this notation, hence the value. 

But, the function signature is right below there, and providing this explicit step may also cause confusion, you are right. Agreed to delete 👍

brozorec (approved):

PR #714, opened 2026-05-20
rwa forced transfer fix

### LINKED ISSUE #713: 🐞 [Bug]: RWA - Forced transfer does not enforce identity verification checks on `to` account
### What happened?

Forced transfer should verify the validity (whitelist) of the `to` account. `from` is not needed.

### Expected behavior

_No response_

ozgunozerk:
Fixes #713 

#### PR Checklist

- [ ] Tests
- [x] Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Added identity verification check for transfer destinations in privileged balance operations, preventing unverified accounts from receiving tokens.

* **Documentation**
  * Clarified that destination addresses require identity verification before accepting transfers, while source addresses do not need verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/stellar-contracts/pull/714?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/tokens/src/rwa/mod.rs
+++ b/packages/tokens/src/rwa/mod.rs
@@ -126,10 +126,15 @@ use crate::fungible::FungibleToken;
 pub trait RWAToken: Pausable + FungibleToken<ContractType = RWA> {
     // ################## CORE TOKEN FUNCTIONS ##################
 
-    /// Forces a transfer of tokens between two whitelisted wallets.
+    /// Forces a transfer of tokens from `from` to `to`.
     /// This function can only be called by the operator with necessary
     /// privileges. RBAC checks are expected to be enforced on the
     /// `operator`.
+    ///
+    /// The `to` address must pass identity verification, but `from` is not
+    /// verified, so this privileged action can still move tokens out of
+    /// accounts whose identity is no longer valid (sanctioned, revoked, or
+    /// compromised wallets).
     ///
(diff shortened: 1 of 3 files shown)


packages/tokens/src/rwa/mod.rs
```diff
@@ -126,11 +126,16 @@ use crate::fungible::FungibleToken;
...
-    /// Forces a transfer of tokens between two whitelisted wallets.
+    /// Forces a transfer of tokens from `from` to `to`.
     /// This function can only be called by the operator with necessary
     /// privileges. RBAC checks are expected to be enforced on the
     /// `operator`.
     ///
+    /// The `to` address must pass identity verification, but `from` is not
+    /// verified, so this privileged action can still move tokens out of
+    /// accounts whose identity is no longer valid (sanctioned, revoked, or
+    /// compromised wallets).
```
brozorec: If `from` has some balance, I believe we can safely assume it has been already verified.

ozgunozerk: but once you are verified, it doesn't mean you will be verified for the future. The wording can be changed as ->

> `from` is not required to be verified

does that work?

brozorec: ok, that works 👍

brozorec (approved):

PR #729, opened 2026-05-26
feat(crypto): Grumpkin affine point arithmetic

### LINKED ISSUE #700: On-chain Grumpkin point arithmetic for Soroban wrapper
## Context

The wrapper contract performs Grumpkin affine point addition / subtraction for every homomorphic balance update: deposit, merge, incoming transfer accumulation (see the internal design doc — link available in the tracking workspace — §7.3, §7.4, §7.6 for the operations and §10.8 for the on-chain arithmetic spec). Soroban natively supports BN254 `Fr` arithmetic (CAP-80), which is exactly Grumpkin's coordinate field, so the wrapper builds Grumpkin point ops on top of `bn254_fr_{add, sub, mul, inv}` host calls.

This is a security-critical primitive: every balance update in the system flows through it. Bugs here are minted tokens. It belongs in a small, fully tested Rust crate that the wrapper depends on — not inlined.

## Scope

Build a `lib/grumpkin/` Rust crate (or contract-local module, depending on deployment layout — see open question) that provides Grumpkin affine point arithmetic on Soroban via `bn254_fr_*` host calls.

- **Encoding**: `Point` as `BytesN<64>` storing `(x, y)` in canonical Fr byte order. Identity `O` is `(0, 0)` and handled as a special case (design doc §5.1, §10.8).
- **`point_add(P1, P2) -> Point`**: standard affine addition. Handle the five cases enumerated in design doc §10.8: left identity, right identity, inverse (`x1 = x2`, `y1 = -y2 mod r`) → returns `O`, doubling (`P1 = P2`), and the generic case. Implements the slope formula in §10.8 via `bn254_fr_{sub, mul, inv}`. The inverse case must be detected and short-circuited before the generic slope formula, otherwise `x1 - x2 = 0` forces a division by zero.
- **Point doubling** (`P1 == P2`): slope `λ_dbl = 3·x1² / (2·y1)` (Grumpkin has `a = 0`; `b = -17` enters only the on-curve check — see §10.8).
- **`point_sub(P1, P2)`**: `P1 + (-P2)` where `-P2 = (x2, -y2 mod r)` (and `-O = O`).
- **Validation**: `is_on_curve(P)` (`y² ≡ x³ - 17 (mod r)`) and `is_not_identity(P)` exported helpers — used by the auditor contract (separate ticket) to validate stored auditor public keys (design doc §3.1, §10.8).
- **Property-based tests**: associativity `(P1 + P2) + P3 == P1 + (P2 + P3)`, commutativity, identity `P + O == P`, inversion `P + (-P) == O`, and Pedersen homomorphism `commit(v1, r1) + commit(v2, r2) == commit(v1+v2, r1+r2)` (random `v`, `r`).
- **Cost benchmarks**: per-op CPU + memory cost, reported in the PR so the wrapper team can size transaction budgets for deposit, merge, and incoming-transfer accumulation.

Out of scope: scalar multiplication (handled inside circuits via Noir `multi_scalar_mul`, not on-chain); the auditor contract; the wrapper integration itself.

## Key Questions
- Should this be a standalone `lib/grumpkin/` crate (reusable, but adds a dependency edge), or a module inside `contracts/wrapper/`? Argument for standalone: the auditor contract also needs `is_on_curve`.
- Does Soroban's `bn254_fr_inv` revert or return an error on zero input? Affects how `point_add` handles the `x1 == x2` collision (must route to doubling or to identity depending on `y` sign).
- Do we need a `multi_add(points: &[Point]) -> Point` batched helper for incoming-transfer accumulation, or is sequential `point_add` fine for v0.8?

## Proposed Approach
Treat this as a self-contained crypto crate with property tests disproportionate to its line count. The cost of a bug is too high to ship on happy-path tests alone.

1. Scaffold `lib/grumpkin/` (or the chosen layout) with no Soroban dependencies in unit tests — back the `bn254_fr_*` ops with a pure-Rust reference implementation so tests run outside the host. The Soroban adapter is a thin trait impl.
2. Implement `point_add` with explicit branches: `P1 == O`, `P2 == O`, `P1 == P2` (doubling), `P1.x == P2.x` and `P1.y == -P2.y` (returns identity), generic case. Each branch gets a named unit test.
3. Implement `point_sub` and validation helpers. Confirm `is_on_curve` against `y² ≡ x³ - 17 (mod r)`.
4. Add property-based tests (`proptest` or equivalent) for the algebraic identities listed in Scope. Run them against both the pure-Rust backend and a Soroban integration harness to ensure host calls give the same answer.
5. Run cost benchmarks under `soroban-cli` test budgets and post the results in the PR description. Flag any single op exceeding the wrapper's per-op CPU envelope.
6. Document the contract: which functions revert, which return errors, and how identity encoding interacts with on-chain storage. The wrapper integration ticket will consume this doc.

## Known References
- Internal design doc — §2.2 (Grumpkin–BN254 cycle), §3.1 (Components — auditor curve check), §5.1 (Balance commitment encoding), §10.7 (CAP-80 dependency), §10.8 (On-Chain Point Arithmetic). Link available in the tracking workspace.
- [CAP-80](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0080.md)
- Soroban host function reference for `bn254_fr_*`

## Acceptance Criteria
- [ ] Implementation matches architecture brief
- [ ] Tests cover main flows and edge cases
- [ ] Review/audit findings addressed (if applicable)
- [ ] PR links back to this issue

brozorec:
## Summary

Adds Grumpkin (y² = x³ − 17 over BN254's scalar field F_r) affine point arithmetic to `packages/contract-utils/src/crypto/`, intended as a building block for on-chain Pedersen commitments, ECDH, and any other Grumpkin-based primitive built on top of CAP-80.

Resolves #700.

- `Grumpkin::{add, sub, neg}` — affine add/sub/neg with full identity handling (left/right identity, inverse, doubling, generic), implemented on top of the BN254 Fr host calls (`bn254_fr_{add, sub, mul, inv}`).
- `Grumpkin::{is_on_curve, is_identity, is_not_identity}` — validation helpers for points entering the system without a soundness proof. `is_on_curve` enforces both the curve equation and the **canonical 32-byte representative** of each coordinate (mirroring the SDK's own `Bn254Fp` validation), so byte equality on validated points is sound for uniqueness / nullifier / commitment-log keys.
- `Point` is a 64-byte `BytesN<64>` laid out as `be(x) || be(y)`; identity is the all-zero encoding.

### Out of scope (per the issue)

- Scalar multiplication — handled in-circuit via Noir's `multi_scalar_mul`.
- Batched `multi_add` — sequential `add` is fine for v0.8; revisit if benchmarks justify.

## Commits

1. `bump soroban-sdk to 26.0.0` — additive: CAP-80 BN254 Fr arithmetic + CAP-78/79/82.
2. `feat(crypto): on-chain Grumpkin affine point arithmetic` — the module + 24 tests.
3. `fix(grumpkin): reject non-canonical coordinate encodings in is_on_curve` — a self-review caught that `Bn254Fr::from_bytes` silently reduces mod r, which would have let two distinct byte encodings of the same logical commitment slip past the validator. Fix + 3 additional tests.

## Test plan

- [x] `cargo +nightly fmt --all -- --check`
- [x] `cargo clippy --release --locked --all-targets -- -D warnings`
- [x] `cargo test --workspace`
- [x] 27 Grumpkin tests pass (16 unit + 11 proptest): each branch (left/right identity, inverse, doubling, generic), neg/sub/identity edges, on-curve / off-curve / non-canonical / identity-encoding rejection, plus the algebraic identities from the issue — commutativity, associativity, identity law, inversion, Pedersen homomorphism. Ground-truth points come from `ark-grumpkin` (dev-dep).

## Notes for reviewers

- **SDK 25 → 26 bump**: required for `Bn254Fr` `Add/Sub/Mul/inv` (no equivalent surface in 25). Bump is additive per the SDK migration notes; the full workspace builds and tests pass unchanged. Worth a glance at the `Cargo.lock` diff in the first commit.
- **No on-curve check inside `add`/`sub`/`neg`**: documented precondition, matches the design's "validate at the trust boundary" stance. The validator is the line of defence; arithmetic is hot-path.
- **Cost benchmarks**: not produced here. Happy to add per-op CPU/memory numbers in a follow-up if maintainers want them in the PR description.

## Summary by CodeRabbit

* **New Features**
  * Added Grumpkin elliptic curve point arithmetic operations for advanced cryptographic support, including point addition, negation, subtraction, identity validation, and on-curve verification.

* **Dependencies**
  * Updated `soroban-sdk` to version 26.0.0.
  * Added `ark-*` cryptographic libraries for curve operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/stellar-contracts/pull/729?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/contract-utils/src/crypto/grumpkin.rs
+++ b/packages/contract-utils/src/crypto/grumpkin.rs
@@ -42,16 +42,28 @@ pub struct Grumpkin;
 
 impl Grumpkin {
     /// Returns the encoded identity element (the point at infinity).
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
     pub fn identity(e: &Env) -> Point {
         BytesN::from_array(e, &[0u8; 64])
     }
 
     /// Returns `true` iff `p` is the identity (all-zero 64-byte encoding).
+    ///
+    /// # Arguments
+    ///
+    /// * `p` - The point to check.
     pub fn is_identity(p: &Point) -> bool {
         p.to_array() == [0u8; 64]
     }
 
     /// Returns `true` iff `p` is not the identity.
+    ///
+    /// # Arguments
+    ///
+    /// * `p` - The point to check.
     pub fn is_not_identity(p: &Point) -> bool {
         !Self::is_identity(p)
     }
@@ -69,6 +81,11 @@ impl Grumpkin {
     /// The identity encoding `(0, 0)` returns `false`: `0² ≠ -17 (mod r)`. To
     /// admit `O` as a valid group element, combine this with
     /// [`Self::is_identity`] at the call site.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p` - The point to validate.
     pub fn is_on_curve(e: &Env, p: &Point) -> bool {
         let bytes = p.to_array();
         if !is_canonical_coord(&bytes[..32]) || !is_canonical_coord(&bytes[32..]) {
@@ -82,6 +99,11 @@ impl Grumpkin {
 
     /// Returns `-p`. For non-identity `p = (x, y)`, this is `(x, -y mod r)`;
     /// `-O = O`.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p` - The point to negate.
     pub fn neg(e: &Env, p: &Point) -> Point {
         if Self::is_identity(p) {
             return p.clone();
@@ -103,6 +125,12 @@ impl Grumpkin {
     ///    two-torsion edge case `y = 0`, since then `y = -y` and `p = -p`.
     /// 4. `p1 = p2` → doubling with slope `λ_dbl = 3·x1² / (2·y1)`.
     /// 5. Otherwise (generic) → slope `λ_add = (y2 - y1) / (x2 - x1)`.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p1` - The first summand.
+    /// * `p2` - The second summand.
     pub fn add(e: &Env, p1: &Point, p2: &Point) -> Point {
         if Self::is_identity(p1) {
             return p2.clone();
@@ -142,11 +170,23 @@ impl Grumpkin {
 
     /// Returns `p1 - p2 = p1 + (-p2)`. Subtraction of a point from itself
     /// returns `O` via the inverse case in [`Self::add`].
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p1` - The minuend.
+    /// * `p2` - The subtrahend.
     pub fn sub(e: &Env, p1: &Point, p2: &Point) -> Point {
         Self::add(e, p1, &Self::neg(e, p2))
     }
 
     /// Encodes `(x, y)` as a 64-byte [`Point`].
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `x` - The x-coordinate as a `Bn254Fr` element.
+    /// * `y` - The y-coordinate as a `Bn254Fr` element.
     pub fn from_xy(e: &Env, x: &Bn254Fr, y: &Bn254Fr) -> Point {
         let mut bytes = [0u8; 64];
         bytes[..32].copy_from_slice(&x.to_bytes().to_array());
@@ -155,6 +195,11 @@ impl Grumpkin {
     }
 
     /// Decodes `(x, y)` from a 64-byte [`Point`].
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p` - The encoded point to decode.
     pub fn coordinates(e: &Env, p: &Point) -> (Bn254Fr, Bn254Fr) {
         let bytes = p.to_array();
         let mut x = [0u8; 32];
--- a/packages/contract-utils/src/crypto/grumpkin.rs
+++ b/packages/contract-utils/src/crypto/grumpkin.rs
@@ -31,6 +31,16 @@ use soroban_sdk::{crypto::bn254::Bn254Fr, BytesN, Env, U256};
 /// infinity).
 pub type Point = BytesN<64>;
 
+/// Canonical byte encoding of the identity element (point at infinity).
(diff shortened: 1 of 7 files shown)


packages/contract-utils/src/crypto/test/grumpkin.rs
```diff
@@ -0,0 +1,434 @@
...
+//! Ground-truth point values are produced by `ark-grumpkin` (off-host) and
+//! piped through the 64-byte `(be(x) || be(y))` encoding into the on-host
+//! arithmetic. The property tests cover the algebraic identities of an
+//! abelian group — associativity, commutativity, identity, inversion — plus
+//! Pedersen homomorphism, and exercise the case-by-case branches of `add`
+//! (left/right identity, inverse, doubling, generic).
+
+extern crate std;
+
+use ark_ec::{AffineRepr, CurveGroup, PrimeGroup};
```
bidzyyys, line 12: Is `ark` audited?

packages/contract-utils/src/crypto/test/mod.rs
bidzyyys, line 1: Looks like an interesting crypto crate. How about making it a Rust crate?

packages/contract-utils/src/crypto/grumpkin.rs
```diff
@@ -0,0 +1,237 @@
...
+        BytesN::from_array(e, &[0u8; 64])
+    }
+
+    /// Returns `true` iff `p` is the identity (all-zero 64-byte encoding).
+    ///
+    /// # Arguments
+    ///
+    /// * `p` - The point to check.
+    pub fn is_identity(p: &Point) -> bool {
+        p.to_array() == [0u8; 64]
```
bidzyyys: Wdyt about making it a const ?

bidzyyys (approved):
LGTM

packages/contract-utils/src/crypto/test/grumpkin.rs
brozorec: I haven't checked but this is used only in tests. We usually don't consider this when using other crates in test env only.

packages/contract-utils/src/crypto/test/mod.rs
brozorec: It can't be used outside of Soroban because it leverages Soroban's host function calls. Once we get it to a release and publish to crates.io, it will become part of the "stellar-contract-utils" crate. I'm not sure whether it's worth it to further push it to a standalone. As it's not a pure rust but Soroban-dependant, I wouldn't advise to make it part of openzeppelin-crypto either.

PR #735, opened 2026-05-29
feat(confidential): wrapper contract + compliance extension

brozorec:
## Summary

Adds the confidential token wrapper — a SEP-41-backed contract with confidential balances and proof-gated transfers — together with a deployer-configurable compliance layer.

The wrapper exposes eleven entry points (register / deposit / merge / withdraw / two transfer variants / four operator-management methods / read accessors) via the `ConfidentialTokenWrapper` trait. Confidential balances are stored as Pedersen commitments on Grumpkin; every state-changing op that consumes private state ships an UltraHonk proof that the wrapper verifies through a separate verifier contract, and every transfer emits dual auditor ciphertexts.

A `Hooks` lifecycle trait runs synchronously at each entry point between auth and storage mutation — the canonical extension point for compliance gates, audit mirroring, rate limiting, etc.

The new `wrapper::compliance` submodule layers four controls (per-account freezing, pluggable external policy, SAC `authorized()` passthrough, unregistered-deposit carve-out) on top of those hooks. It ships a turnkey `ComplianceHooks` (wire as `type Hooks = ComplianceHooks;`) and an admin trait `ConfidentialCompliance` for freeze/unfreeze/config rotation.

## What's in this PR

- **`packages/contract-utils/src/crypto/grumpkin.rs`** — adds scalar multiplication (`Grumpkin::mul`) on the affine API, with tests, needed by the wrapper's deposit commitment.
- **`packages/tokens/src/confidential/wrapper/`** — the new module:
  - `mod.rs` — `ConfidentialTokenWrapper` trait, `Hooks` trait + `NoHooks`, all errors / events / payload types.
  - `storage.rs` — `*_no_auth` orchestration helpers, public-input assembly per DESIGN §7, cross-contract verifier/auditor calls.
  - `test.rs` — wrapper-level unit tests.
- **`packages/tokens/src/confidential/wrapper/compliance/`** — the compliance submodule:
  - `mod.rs` — `Policy` interface, `ComplianceError`, events, `ComplianceHooks`, `ConfidentialCompliance` trait.
  - `storage.rs` — `ComplianceConfig`, storage keys, `*_no_auth` admin helpers, gating primitives (`gate_account` / `check_policy` / `check_sac`).
  - `test.rs` — 29 tests covering the no-config short-circuit, freeze flow, policy gate, SAC passthrough, unregistered-deposit carve-out, config rotation, the `ConfidentialCompliance` trait via its generated client, and end-to-end hook dispatch through the wrapper.
- **`docs`** — module-level `## Underlying Token Requirements` section documenting the exact-transfer-semantics requirement on the wrapped SEP-41 token (SAC or OZ `fungible` are conformant); `deposit_no_auth` / `withdraw_no_auth` carry matching `# Notes` blocks.

## Design choices worth reviewing

- **Per-account freezing in the wrapper, not the SAC.** Freezes a confidential account directly via persistent storage so the wrapper's gates fire on register / transfer / withdraw without round-tripping through the SAC. SAC passthrough remains opt-in via `sac_passthrough`.
- **Pluggable `Policy` interface.** A single registry contract can serve multiple wrappers — `is_authorized(account, wrapper)` receives the wrapper's own address.
- **Unregistered-deposit carve-out** (COMPLIANCE.md §4). When `permit_unregistered_deposit` is set, `on_deposit` skips freeze + policy on a depositor with no confidential account, but still runs SAC passthrough (since the underlying SEP-41 transfer would fail anyway if the SAC has unauthorized them).
- **Admin trait setters have no default body.** `freeze` / `unfreeze` / `set_compliance_config` accept `admin: Address` and require the contract author to override — a default would either double-auth under `#[only_owner]` / `#[only_role]` or leave the entry point ungated.
- **Storage TTL convention.** Reads on `Frozen(account)` extend TTL; writes don't (per CLAUDE.md). Compliance config sits in instance storage.
- **Out of scope:** §5 (pooled-custody clawback / seizure) is intentionally deferred.

## Test plan

- [x] `cargo test --package stellar-tokens` — 607 passing (29 new in compliance).
- [x] `cargo +nightly fmt --all -- --check` — clean.
- [x] `cargo clippy --release --locked --package stellar-tokens -- -D warnings` — clean.
- [ ] WASM release build per-package
- [ ] Coverage threshold (`cargo llvm-cov --workspace --fail-under-lines 90`)
- [ ] Manual review of: gate sequencing in `ComplianceHooks::on_deposit` (carve-out branch), public-input order in `storage.rs` against DESIGN §7, the "no default body" rationale in the `ConfidentialCompliance` trait docstring

## Notes for reviewers

- The wrapper depends on `ConfidentialVerifier::verify_proof`, whose UltraHonk backend is still unaudited — module docstring already carries the "Not Production Ready" warning.
- The exact-transfer-semantics requirement is currently documented rather than enforced at runtime. A defense-in-depth alternative would be to measure the wrapper's own balance pre/post `token.transfer` and credit/debit by the delta. Happy to swap docs → runtime measurement if preferred.

## Summary by CodeRabbit

* **New Features**
  * Added Grumpkin cryptographic operations for elliptic curve calculations
  * Introduced confidential token wrapper enabling private transactions with homomorphic commitments
  * Added compliance framework supporting account freezing and external policy authorization

* **Tests**
  * Added comprehensive test suites for cryptographic operations, wrapper functionality, and compliance features

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/stellar-contracts/pull/735?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/tokens/src/confidential/wrapper/mod.rs
bidzyyys, line 1: Shouldn't you hide it under feature flag? (Still unaudited backend).

bidzyyys (approved):
Looks good.

packages/tokens/src/confidential/wrapper/compliance/mod.rs
```diff
@@ -0,0 +1,353 @@
+//! # Confidential Wrapper Compliance Extension
+//!
+//! Deployer-configurable controls layered on top of the
+//! [`ConfidentialTokenWrapper`](super::ConfidentialTokenWrapper): per-account
```
ozgunozerk: we decided to not use `super` keyword across the library. I'd prefer to keep it like it

packages/contract-utils/src/crypto/grumpkin.rs
```diff
@@ -190,6 +222,38 @@ impl Grumpkin {
...
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `p` - The base point.
+    /// * `scalar` - The scalar multiplier. Sized for the on-chain consumers'
+    ///   needs (token amounts and similar bounded quantities); off-chain
+    ///   provers typically use larger scalars and should pre-compute on the
+    ///   prover side rather than calling this function.
+    pub fn mul(e: &Env, p: &Point, scalar: u128) -> Point {
+        if scalar == 0 || Self::is_identity(p) {
```
ozgunozerk, line 239: This is super confusing for me, and took me 10 minutes to understand. I found out the reason:

`identity` is the identity element w.r.t to group addition operation on elliptic curves, it's not the `identity` for `mul` operation. 

But for someone reading this operation, I think they would expect `identity` to be `1`, because we are in `mul` function's definition. Then, this code doesn't make sense. 

I suggest either of these:
- rename `identity` related naming to `additive_identity`, or
- put an explanatory comment here and other places where identity is used in another setting than addition

packages/tokens/src/confidential/compliance/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+    ///
+    /// * topics - `["frozen", account: Address]`
+    /// * data - `[]`
+    ///
+    /// # Security Warning
+    ///
+    /// Implementations MUST authorize `admin` before calling
+    /// [`storage::freeze_no_auth`]. The trait cannot provide a default body
+    /// — see the trait-level docstring for the rationale.
+    fn freeze(e: &Env, account: Address, admin: Address);
```
ozgunozerk: should we rename `admin` to `operator`?

packages/tokens/src/confidential/wrapper/compliance/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+/// 2. When `config.policy = Some(p)`, calls `p.is_authorized` and reverts
+///    [`ComplianceError::NotAuthorizedByPolicy`] on `false`.
+/// 3. When `config.sac_passthrough = true`, calls the underlying SEP-41 token's
+///    `authorized` view and reverts [`ComplianceError::NotAuthorizedBySac`] on
+///    `false`.
+///
+/// Special cases:
+///
+/// * [`on_register`](Hooks::on_register) skips the freeze branch (no slot
+///   exists yet) but still applies policy and SAC.
```
ozgunozerk: what does `slot` mean here?

packages/tokens/src/confidential/wrapper/compliance/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+        e: &Env,
+        operator: &Address,
+        from: &Address,
+        to: &Address,
+        _payload: Val,
+    ) {
+        let Some(config) = storage::compliance_config(e) else {
+            return;
+        };
+        storage::gate_account(e, operator, &config);
```
ozgunozerk: i don't think operator should be gated here. We followed a similar approach in RWA as well. Operator can be frozen, but should still be able to invoke the transfer, because it is not their balance which is getting affected imo.

packages/tokens/src/confidential/wrapper/compliance/mod.rs
```diff
@@ -0,0 +1,353 @@
...
+    Unfrozen { account: account.clone() }.publish(e);
+}
+
+/// Event emitted when the compliance configuration is set or rotated.
+#[contractevent]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ComplianceConfigChanged {
+    pub policy: Option<Address>,
+    pub sac_passthrough: bool,
+    pub permit_unregistered_deposit: bool,
```
ozgunozerk: those 2 concepts are still aren't crystal clear to me, commenting here just as a reminder to myself

packages/tokens/src/confidential/wrapper/mod.rs
```diff
@@ -0,0 +1,877 @@
...
+//! fee-on-transfer, rebasing, or otherwise malicious token implementation,
+//! the confidential ledger would drift from the on-chain reserves —
+//! credit a higher amount than was actually received, or pay out less than
+//! was debited.
+//!
+//! Supported token implementations:
+//!
+//! - the Stellar Asset Contract (SAC), and
+//! - OpenZeppelin's [`fungible`](crate::fungible) token (which preserves SEP-41
+//!   exact-transfer semantics).
```
ozgunozerk: there can/will be other tokens that can be compliant. Such a wording excludes other possibilities. I think we should rephrase here, in such a way:

Tokens that are SEP-41 compliant are supported, for example:
- openzeppelin
- SAC

packages/tokens/src/confidential/wrapper/mod.rs
```diff
@@ -0,0 +1,877 @@
...
+//!
+//! Deploying the wrapper over any other token implementation is the
+//! deployer's responsibility — verify that `transfer` does not skim a fee
+//! or otherwise diverge from exact-transfer semantics before doing so.
+//!
+//! ## Storage
+//!
+//! Singleton configuration (`token`, `verifier`, `auditor`, `wrap`) lives in
+//! instance storage. Per-account state lives in persistent storage; reads
+//! extend TTL, writes do not (CLAUDE.md storage convention).
```
ozgunozerk: this part should be removed imo

ozgunozerk (changes requested):
I think we don't need the `no_auth` suffix for the storage functions. For the last modules, we quit that convention, and we only have them when there is the auth and no_auth version both present in the storage file. 

This change will make the function names shorter

packages/tokens/src/confidential/wrapper/mod.rs
```diff
@@ -0,0 +1,877 @@
...
+            &from,
+            &to,
+            &decoded.payload,
+            &decoded.proof,
+        );
+    }
+
+    /// Escrows an allowance from `account`'s spendable balance into a
+    /// `(account, operator)` delegation slot. Reverts if a delegation already
+    /// exists for the pair (single-slot semantics).
```
ozgunozerk: again, this slot terminology is confusing for me. Can we expand on it, or use a different wording for it?

packages/tokens/src/confidential/wrapper/mod.rs
```diff
@@ -0,0 +1,877 @@
...
+            &operator,
+            expiration_ledger,
+            &decoded.payload,
+            &decoded.proof,
+        );
+    }
+
+    /// Revokes the `(account, operator)` delegation and folds the
+    /// remaining escrowed allowance back into `account`'s spendable balance
+    ///. Works for both active and expired-but-not-revoked
```
ozgunozerk: the `.` in the beginning of the line looks weird

packages/tokens/src/confidential/wrapper/mod.rs
```diff
@@ -0,0 +1,877 @@
...
+#[repr(u32)]
+pub enum WrapperError {
+    /// Indicates `account` already has a confidential account registered.
+    AccountAlreadyRegistered = 3500,
+    /// Indicates the target account is not registered.
+    AccountNotRegistered = 3501,
+    /// Indicates a public amount argument is negative.
+    NegativeAmount = 3502,
+    /// Indicates a delegation already exists for `(account, operator)`
+    ///.
```
ozgunozerk: we can carry the dot upwards, no need for a new line just for a dot

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+    /// Confidential verifier contract used for `verify_proof`. Instance
+    /// storage.
+    Verifier,
+    /// Confidential auditor contract used for `get_key`. Instance storage.
+    Auditor,
+    /// `wrap = address_to_field(env.current_contract_address())` (DESIGN §2.7,
+    /// §3.5), bound into every owner-initiated circuit's `vk` derivation.
+    /// Stored as a canonical 32-byte big-endian `Bn254Fr` representative.
+    /// Instance storage.
+    Wrap,
```
ozgunozerk: this part is also confusing, even with inline docs

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+pub struct WithdrawPayload {
+    pub c_spend_new: Point,
+    pub b_tilde: BytesN<32>,
+    pub r_e: Point,
+    pub sigma: BytesN<32>,
+    pub b_aud_s: BytesN<32>,
+}
+
+/// Envelope decoded from the `data: Bytes` argument of
+/// [`super::ConfidentialTokenWrapper::withdraw`].
```
ozgunozerk: `super` keyword is in many places, would be great to replace them with the full path or if the relevant items are imported, using the imported path

packages/tokens/src/confidential/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+///
+/// # Errors
+///
+/// * [`WrapperError::WrapNotSet`] - When the wrapper has not been constructed.
+pub fn get_wrap(e: &Env) -> BytesN<32> {
+    e.storage()
+        .instance()
```
ozgunozerk, line 294: since I didn't fully grasp what `wrap` is, I cannot give an opinion here, but just raising a question, should this be instance or persistent storage? Feel free to resolve this convo on your own

packages/tokens/src/confidential/compliance/mod.rs
brozorec: What about renaming it to `caller` because `operator` in the CT context has another meaning (ref. `confidential_transfer_from`)?

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+pub fn get_wrap(e: &Env) -> BytesN<32> {
+    e.storage()
+        .instance()
+        .get(&WrapperStorageKey::Wrap)
+        .unwrap_or_else(|| panic_with_error!(e, WrapperError::WrapNotSet))
+}
+
+/// Returns the [`ConfidentialAccount`] stored under `account`.
+///
+/// Extends the entry's persistent TTL on a hit.
```
ozgunozerk: I don't think we need to note TTL extension on a hit in inline docs. Saw this in many places.

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+pub fn get_account(e: &Env, account: &Address) -> ConfidentialAccount {
+    get_persistent_entry::<ConfidentialAccount>(e, &WrapperStorageKey::Account(account.clone()))
+        .unwrap_or_else(|| panic_with_error!(e, WrapperError::AccountNotRegistered))
+}
+
+/// Returns the [`OperatorDelegation`] stored under `(owner, operator)`.
+///
+/// Extends the entry's persistent TTL on a hit. Does not apply the expiry
+/// filter; callers that want the spending-authority view should use
+/// [`is_operator`] instead.
```
ozgunozerk: What is expiry filter? Is it expiry for TTL? Or is it something else? 
What is spending-authority and the relationship with `is_operator`? 

This is again cryptic imo

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+pub fn get_delegation(e: &Env, owner: &Address, operator: &Address) -> OperatorDelegation {
+    get_persistent_entry::<OperatorDelegation>(
+        e,
+        &WrapperStorageKey::Delegation(owner.clone(), operator.clone()),
+    )
+    .unwrap_or_else(|| panic_with_error!(e, WrapperError::DelegationNotFound))
+}
+
+/// Returns `true` iff a delegation exists for `(owner, operator)` and the
+/// current ledger sequence does not exceed its `expiration_ledger`.
```
ozgunozerk: I believe the only diff between is_operator and `get_delegation` is the expiry check on the ledger. 

Do we need `get_delegation` in this case? 

And, I believe the naming for these is confusing. The explanation of `is_operator` is beginning with:
```
Returns `true` iff a delegation exists for `(owner, operator)`
```

Yet, the there is another function with the name `get_delegation`. I suggest renaming these functions accordingly, or getting rid of one

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+}
+
+/// Returns whether an account is registered, without panicking.
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `account` - The owner address.
+pub fn account_exists(e: &Env, account: &Address) -> bool {
+    e.storage().persistent().has(&WrapperStorageKey::Account(account.clone()))
```
ozgunozerk: why do we not extend persistent storage up there in other functions, but not here and below?

packages/tokens/src/confidential/wrapper/storage.rs
brozorec: `Wrap` is the address of the wrapper stored as a Bn254Fr Field element, calculated and stored in instance storage once so that we don't need to recalculate every time we need it.

packages/tokens/src/confidential/compliance/mod.rs
ozgunozerk: but we use operator for the very same meaning in all the other contracts. I think that is also an important factor. 

Why don't we name `operator` in the transfer_from as `spender`? We have it like that in the fungible token?

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+/// resets the receiving slot to the identity.
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `account` - The owner address.
+///
+/// # Errors
+///
+/// * [`WrapperError::AccountNotRegistered`] - When `account` is not registered.
```
ozgunozerk: should be `refer to get_account errors` instead

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+/// * `to` - The SEP-41 recipient.
+/// * `amount` - The non-negative withdrawal amount.
+/// * `payload` - The decoded [`WithdrawPayload`].
+/// * `proof` - The raw UltraHonk proof bytes.
+///
+/// # Errors
+///
+/// * [`WrapperError::NegativeAmount`] - When `amount < 0`.
+/// * [`WrapperError::AccountNotRegistered`] - When `from` is not registered.
+/// * [`WrapperError::InvalidProof`] - When the proof fails verification.
```
ozgunozerk: these errors also not present directly in the function. We should probably have the LLM skim through all the code, and keep only the direct ones, and use the `refer to` convention for the other errors.

packages/tokens/src/confidential/wrapper/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+/// Intended for use in `__constructor` against
+/// `env.current_contract_address()` and then stored via [`set_wrap`].
+///
+/// # Arguments
+///
+/// * `e` - Access to the Soroban environment.
+/// * `addr` - The address to compress.
+pub fn compute_wrap(e: &Env, addr: &Address) -> BytesN<32> {
+    address_to_field(e, addr)
+}
```
ozgunozerk: what is the use for this function? Just use`address_to_field` directly?

packages/tokens/src/confidential/storage.rs
```diff
@@ -0,0 +1,1363 @@
...
+/// * `owner` - The delegating account.
+/// * `operator` - The delegated operator.
+///
+/// # Errors
+///
+/// * [`WrapperError::DelegationNotFound`] - When no delegation exists for the
+///   pair.
+fn delete_delegation(e: &Env, owner: &Address, operator: &Address) {
+    let key = WrapperStorageKey::Delegation(owner.clone(), operator.clone());
+    if !e.storage().persistent().has(&key) {
```
ozgunozerk, line 1269: we don't extend persistent on read on here, but we do in below, why?

packages/tokens/src/confidential/wrapper/storage.rs
brozorec: For `get_account` errors, I prefer to leave that explicit, because for `deposit()` we check only `to`, while for `confidential_transfer` we check both `from` and `to`. Being explicit here is worth it imo, because it provides more clarity.

packages/tokens/src/confidential/wrapper/storage.rs
brozorec: good point 👍

packages/tokens/src/confidential/wrapper/storage.rs
brozorec: expiry is about `expiration_ledger` in `OperatorDelegation`

brozorec: I agree, leaving just "Returns the [`OperatorDelegation`] stored under `(owner, operator)`." is way better.

packages/tokens/src/confidential/storage.rs
brozorec: that's because we are about to delete it here

packages/tokens/src/confidential/wrapper/storage.rs
brozorec: renamed the functions, lmk whether this is clearer

packages/contract-utils/src/crypto/grumpkin.rs
brozorec: > `identity` is the identity element w.r.t to group addition operation on elliptic curves

Not only in additions. The identity element is the point at infinity and when multiplying it with a scalar, we get back the identity.

packages/tokens/src/confidential/wrapper/mod.rs
brozorec: @ozgunozerk wdyt?

packages/tokens/src/confidential/wrapper/storage.rs
ozgunozerk: i don't see much value in being explicit here compared to having it following the convention, but don't have a strong stance, you can choose whatever you'd prefer, it's all good!

packages/tokens/src/confidential/compliance/mod.rs
brozorec: because we want to stay closer to the semantics in https://eips.ethereum.org/EIPS/eip-7984

packages/contract-utils/src/crypto/grumpkin.rs
ozgunozerk: That's exactly my point. It's not identity in multiplication. 

Number x Identity = Number

That's the definition of Identity. 

In the case you mention

Number x Identity = Identity 

This is not a property of identity. 

The identity properties for that element is only preserved in addition from what I've seen.

packages/tokens/src/confidential/compliance/mod.rs
ozgunozerk: but they also used the term `spender` here: 
```
function isOperator(address holder, address spender) external view returns (bool)
```

I don't think `operator` wins over `spender` here for the eip 7984 semantics, in fact, I'd say spender wins

packages/tokens/src/confidential/docs/COMPLIANCE.github.md
```diff
@@ -2,7 +2,7 @@
 
 ## Abstract
 
-This document specifies optional, deployer-configurable controls layered on top of the core Confidential Token Wrapper (see [DESIGN.md](DESIGN.md)). It covers account freezing, SAC authorization passthrough, pluggable authorization policies, unregistered-deposit handling, and the pooled-custody clawback flow.
+This document specifies optional, deployer-configurable controls layered on top of the core Confidential Token Wrapper (see [DESIGN.md](DESIGN.md)). It covers account freezing, SAC authorization passthrough, pluggable authorization policies, customization patterns for the `Hooks` extension surface, and the pooled-custody clawback flow.
 
 All controls are configured at construction time through a single `compliance: Option<ComplianceConfig>` entry. A vanilla deployment leaves the entry empty and pays no compliance overhead. Regulated deployments populate the slot once; subsequent state changes (freeze toggles, admin rotation, policy swap) flow through admin-gated entry points.
```
ozgunozerk: there is still `slot` and `wrapper` in this document. Could be other remnants in the codebase as well.

packages/tokens/src/confidential/compliance/storage.rs
```diff
@@ -0,0 +1,232 @@
...
+};
+
+// ################## TYPES ##################
+
+/// Compliance configuration written once at construction and rotatable under
+/// admin auth thereafter. Stored as an instance storage entry.
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ComplianceConfig {
+    /// Optional external authorization policy (see [`super::Policy`]). `None`
```
ozgunozerk: There are still a lot of `super` occurrences in the code

ozgunozerk (changes requested):
I think the word `wrap` should be removed/reworked within this PR, because we decided to rename the module, it feels like a remnant

packages/tokens/src/confidential/compliance/mod.rs
brozorec: I'm also more inclined to using `spender`, but wanted to stay closer to 7984, knowing that clashes with the library conventions. But your pushback makes me reconsider this...

> they also used the term spender here

Hmm, I assume the usage of both terms is an oversight tbh, it doesn't make sense to use `operator` everywhere except in this specific function.

ozgunozerk: I don't think this difference is a crucial point for staying closer to 7984. They used the term interchangeably themselves. And considering our library, and considering the meaning of the words, `spender` makes more sense as a name imo

ozgunozerk (approved):
Big chunk of work! 🥳

PR #745, opened 2026-06-03
time limit transfer compliance module

### LINKED ISSUE #742: 🏗️ [Core Feature]: Time Limit Transfer Modules

Compliance module for time based transfer limits

ozgunozerk:
Fixes #742 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added time-window transfer limit compliance module for enforcing cumulative outgoing transfer caps over rolling periods
  * Supports batch operations for configuring and removing limits
  * Enables identity-based transfer aggregation across multiple wallets
  * Automatic counter reset when time windows expire

* **Tests**
  * Comprehensive test coverage for time-window transfer limits functionality

### DIFF
--- a/packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
+++ b/packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
@@ -9,20 +9,20 @@ use crate::rwa::compliance::modules::{
 };
 
 /// A single time-window limit configured for a token: at most `limit_value`
-/// tokens may be sent within a window lasting `limit_duration` seconds. A
+/// tokens may be sent within a window lasting `limit_duration` ledgers. A
 /// window opens with the first transfer after the previous one elapsed;
 /// per-identity consumption against the cap is tracked by
 /// [`TransferCounter`].
 #[contracttype]
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct TransferLimit {
-    pub limit_duration: u64,
+    pub limit_duration: u32,
     pub limit_value: i128,
 }
 
 /// The cumulative volume one identity has sent within its currently active
 /// window: `value` accumulates against the matching [`TransferLimit`]'s cap
-/// until the ledger timestamp reaches `deadline` (the moment the window
+/// until the ledger sequence reaches `deadline` (the moment the window
 /// ends), after which the next transfer restarts the counter for a fresh
 /// window.
 ///
@@ -33,7 +33,7 @@ pub struct TransferLimit {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct TransferCounter {
     pub value: i128,
-    pub deadline: u64,
+    pub deadline: u32,
 }
 
 /// Storage key fields for a per-(token, identity, window) counter entry.
@@ -42,7 +42,7 @@ pub struct TransferCounter {
 pub struct TransferCounterKey {
     pub token: Address,
     pub identity: Address,
-    pub limit_duration: u64,
+    pub limit_duration: u32,
 }
 
 #[contracttype]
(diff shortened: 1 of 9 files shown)


packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
```diff
@@ -0,0 +1,363 @@
...
+///
+/// # Arguments
+///
+/// * `_e` - Access to the Soroban environment.
+/// * `_to` - The recipient address.
+/// * `_amount` - The minted amount.
+/// * `_token` - The token address.
+pub fn can_create(_e: &Env, _to: &Address, _amount: i128, _token: &Address) -> bool {
+    true
+}
```
brozorec: this seems redundant

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
```diff
@@ -0,0 +1,363 @@
...
+/// tokens may be sent within a window lasting `limit_duration` seconds. A
+/// window opens with the first transfer after the previous one elapsed;
+/// per-identity consumption against the cap is tracked by
+/// [`TransferCounter`].
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TransferLimit {
+    pub limit_duration: u64,
+    pub limit_value: i128,
+}
```
brozorec, line 21: I guess we use seconds because of the sol example, but shouldn't we better stick with the codebase convention about using ledgers for measuring time?

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// The cumulative volume one identity has sent within its currently active
+/// window: `value` accumulates against the matching [`TransferLimit`]'s cap
+/// until the ledger timestamp reaches `deadline` (the moment the window
+/// ends), after which the next transfer restarts the counter for a fresh
+/// window.
+///
+/// An identity will will/may have multiple active counters at once if
```
brozorec:
```suggestion
/// An identity will/may have multiple active counters at once if
```

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
```diff
@@ -0,0 +1,363 @@
...
+/// # Events
+///
+/// * topics - `["time_transfer_limit_removed", token: Address]`
+/// * data - `[limit_duration: u64]`
+///
+/// # Notes
+///
+/// Counters recorded for the removed window are not erased. If the same
+/// window duration is configured again later, an unexpired counter resumes
+/// counting where it left off.
```
brozorec, line 249: What's the reasoning for not clearing the state after a removal?

brozorec (changes requested):

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
ozgunozerk: hmm, we need to have `can_create` in an explicit way, but we can definitely override that in the example contract as well. I agree having a storage function just to have `true` is a bit superfluous 👍

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
ozgunozerk: i thought about that. The reason I went with seconds is, this will be probably adopted by tradFi, and how to convert week or days to ledgers? This will be another friction point. Hence I wanted to go with seconds. But your argument here is also strong. We can discuss this further, I don't have a strong stance, I just want to optimize this for the tradFi

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
ozgunozerk: Let me double check this with a fresh mind tomorrow to be sure I'm not missing anything. Good catch!

ozgunozerk: Actually my reasoning was, due to technical infeasibility. The Counters are registered with `(token, identity, limit_duration)` key, and removing that would require us to iterate through each identity that is present for this token, not a great idea.

Then, I found out Tokeny is following the same approach with us here. 

And the counters will get expired in the next window nevertheless, so I think it wouldn't be a big difference whether to keep or remove the counters tbh.

packages/tokens/src/rwa/compliance/modules/time_transfers_limits/storage.rs
brozorec: Another argument against diverging from the convention is that the rwa itself already uses ledgers, eg in `transfer_from`. I believe it's worth staying consistent. Not super convinced about the friction point: if all other users can deal with ledgers, why tradfi won't be able to :)

brozorec (approved):
Approving this PR, but I'd recommend switching from seconds to ledgers

PR #746, opened 2026-06-04
fix(confidential): event emission and non-canonical encodings

brozorec:
## Summary by CodeRabbit

* **New Features**
  * Added validation for canonical encoding of cryptographic field elements and Grumpkin points in the confidential token system.
  * New error code (3514) introduced for non-canonical encoding rejections.

* **Bug Fixes**
  * Contract now enforces canonical encoding checks at verifier boundaries to prevent byte-ambiguity issues.

* **Documentation**
  * Added comprehensive guide for the confidential token suite.
  * Updated design documentation explaining encoding validation requirements and trust boundaries.

### DIFF
--- a/packages/tokens/src/confidential/README.md
+++ b/packages/tokens/src/confidential/README.md
@@ -20,7 +20,7 @@ real value.
 
 ## Layout
 
-```
+```text
 confidential/
 ├── mod.rs          # ConfidentialToken trait + Hooks + events
 ├── storage.rs      # operation-level orchestration, public-input assembly
--- a/packages/tokens/src/confidential/README.md
+++ b/packages/tokens/src/confidential/README.md
@@ -35,7 +35,7 @@ confidential/
 
 A single deployment is **three contracts** wired together:
 
-1. A `ConfidentialToken` contract (wraps one SEP-41 asset).
+1. A `ConfidentialToken` contract.
 2. A `ConfidentialAuditor` registry (Grumpkin public keys, indexed by
    `auditor_id`; reusable across multiple tokens).
 3. A `ConfidentialVerifier` registry (one UltraHonk VK per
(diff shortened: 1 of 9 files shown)


packages/tokens/src/confidential/README.md
```diff
@@ -0,0 +1,100 @@
...
+├── compliance/     # turnkey ComplianceHooks: freeze, SAC passthrough, external policy
+├── circuits/       # Noir/UltraHonk circuits + pinned VKs
+└── docs/
+    ├── DESIGN.md       # full protocol specification
+    └── COMPLIANCE.md   # compliance-extension specification
+```
+
+A single deployment is **three contracts** wired together:
+
+1. A `ConfidentialToken` contract (wraps one SEP-41 asset).
```
ozgunozerk: I still think `wrap` is confusing due to it signals the underlying contract will be in play in every operation

packages/tokens/src/confidential/storage.rs
```diff
@@ -555,6 +557,8 @@ pub fn merge(e: &Env, account: &Address) {
 /// * [`ConfidentialTokenError::NegativeAmount`] - When `amount < 0`.
 /// * [`ConfidentialTokenError::AccountNotRegistered`] - When `from` is not
 ///   registered.
+/// * [`ConfidentialTokenError::NonCanonicalEncoding`] - When point coordinates
+///   or fields from the proof are non-canonical.
```
ozgunozerk, line 561: this error is added to almost every function in this file. However, the function body is not changed. I assume this error will originate in `append_point` only. Shouldn't we use the `refer to` notation instead? Is there a specific reason to deviate from the convention and make this one explicit?

ozgunozerk (approved):
Readme is great!

I'm approving in advance to unblock you. Already made my comments, feel free to address them in a way you see fit, and then merge 🥳

packages/tokens/src/confidential/storage.rs
brozorec: The reason to add those errors explicitly is that `append_point` and `append_field` are not public, so not rendered in the generated docs.

PR #747, opened 2026-06-05
transfer restrict

### LINKED ISSUE #744: 🏗️ [Core Feature]: Transfer Restrict Module

Compliance module for restricting transfers

ozgunozerk:
Fixes #744 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Release Notes

* **New Features**
  * Transfer restriction module: Create per-token allowlists to control which users can transfer tokens.
  * Time-based transfer limits: Transfer restrictions now enforce windows based on time duration rather than block counts.

* **Improvements**
  * Duration parameters updated to support extended time ranges in transfer limit configurations.


packages/tokens/src/rwa/compliance/modules/transfer_restrict/mod.rs
```diff
@@ -0,0 +1,199 @@
+//! Transfer-restriction compliance module — Stellar port of T-REX
+//! [`TransferRestrictModule.sol`][trex-src].
+//!
+//! Maintains a per-token address allowlist. A transfer passes when the
+//! sender is allowlisted; otherwise the recipient must be allowlisted.
```
brozorec: Can we reformulate this sentence? It sounds confusing. The intention is that either sender, or recipient should be allowlisted, right?

packages/tokens/src/rwa/compliance/modules/transfer_allow/mod.rs
```diff
@@ -0,0 +1,199 @@
+//! Transfer-restriction compliance module — Stellar port of T-REX
+//! [`TransferRestrictModule.sol`][trex-src].
```
brozorec, line 2: The name `transfer_restrict` sounds misleading to me, shouldn't it be better the opposite, ie. `transfer_allow`?

packages/tokens/src/rwa/compliance/modules/transfer_restrict/storage.rs
```diff
@@ -0,0 +1,191 @@
...
+/// * `user` - The address to disallow.
+///
+/// # Security Warning
+///
+/// This helper performs no authorization checks.
+pub fn remove_user_allowed(e: &Env, token: &Address, user: &Address) {
+    e.storage()
+        .persistent()
+        .remove(&TransferRestrictStorageKey::AllowedUser(token.clone(), user.clone()));
+}
```
brozorec: While still keeping them public, I'd suggest grouping those two functions as low-level helpers.

packages/tokens/src/rwa/compliance/modules/transfer_allow/mod.rs
brozorec: We are following this one https://github.com/TokenySolutions/T-REX/blob/main/contracts/compliance/modular/modules/TransferRestrictModule.sol but still think there's a name mismatch.

ozgunozerk: I agree, the naming you suggest would be better 👍

brozorec (approved):

PR #760, opened 2026-06-26
max balance fail on recovery fix

### LINKED ISSUE #757: 🐞 [Bug]: RWA Max Balance cannot find the deleted account on `recovery` path
### What happened?

On recovery, the old address is deleted, then the compliance module checks the for the `from` address through IRS, and cannot find it, and reverts

### Expected behavior

_No response_

ozgunozerk:
Fixes #757 

#### PR Checklist

- [ ] Tests
- [ ] Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transfer handling for recovered accounts so balances resolve correctly after identity recovery.
  * Forced transfers now correctly treat same-identity cases as no-ops instead of reverting.
  * Transfers to third parties after recovery now update account limits and bookkeeping using the resolved identities.
* **Tests**
  * Added regression coverage for forced transfers after identity recovery, including same-identity and third-party scenarios.

### DIFF
--- a/packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
+++ b/packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
@@ -231,6 +231,12 @@ pub fn mark_preset_completed(e: &Env, token: &Address) {
 /// where `from` and `to` resolve to the same identity are no-ops: no debit,
 /// no credit, no cap check.
 ///
+/// When `from`'s live identity entry has already been removed by account
+/// recovery, the sender is resolved through the IRS recovery record rather than
+/// reverting on the source lookup, so a forced/recovery transfer still updates
+/// the books. Once `from` recovered to `to`, both sides resolve to the same
+/// identity and the movement is a same-identity no-op.
+///
 /// # Arguments
 ///
 /// * `e` - Access to the Soroban environment.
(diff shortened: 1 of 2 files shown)


packages/tokens/src/rwa/compliance/modules/max_balance/storage.rs
```diff
@@ -266,9 +272,23 @@ pub fn on_transfer(
...
+    // instead of reverting here. A sender that was never registered (no live
+    // identity and no recovery record) still reverts via the final
+    // `stored_identity`, exactly as before.
+    let id_from = match irs.try_stored_identity(from) {
+        Ok(Ok(id)) => id,
+        _ => match irs.get_recovered_to(from) {
+            Some(recovered_wallet) => irs.stored_identity(&recovered_wallet),
+            None => irs.stored_identity(from),
+        },
+    };
```
brozorec, line 292: Although achieving the same outcome, what about capturing the error from `IRS` and panicking with it instead of directly re-calling the function. 

Well, we keep `irs.stored_identity(from)` for the unmatched cases, but that's more like a  unreachable kind of termination.
```
let id_from = match irs.try_stored_identity(from) {
    Ok(Ok(id)) => id,
    Err(Ok(err)) => match irs.get_recovered_to(from) {
        Some(recovered_wallet) => irs.stored_identity(&recovered_wallet),
        None => panic_with_error!(e, err),
    },
    _ => irs.stored_identity(from),
};
```

brozorec (approved):

PR #792, opened 2026-07-13
Derive the withdraw auditor checkpoint pad from the second sponge squeeze (audit N-07)

### LINKED ISSUE #785: [Audit N-07] Withdraw auditor checkpoint pad collides with transfer amount pad
`encrypt_auditor_sender_balance` pads the withdraw balance checkpoint with `Poseidon2(AUDITOR_SENDER, [s_x, sigma])` — the *first* sponge output, which is also the transfer circuit's amount pad. `transfer`, `set_spender`, and `revoke_spender` all use the *second* output for balance pads; only withdraw diverges. Under `(r_e, sigma)` reuse, anyone who knows a transfer amount (e.g. its recipient) can recover the pad from the published `v_tilde_aud_s` and decrypt a later withdraw's balance checkpoint.

**Planned fix:** derive the checkpoint from `sponge_squeeze_2(AUDITOR_SENDER, s_x, sigma)[1]`, restoring the "balance pad = second output" invariant across all circuits → regenerate the withdraw VK + testdata; update DESIGN W_a3/W_a4 and document the unique-`(r_e, sigma)`-per-operation requirement.

Files: `packages/tokens/src/confidential/circuits/lib/src/lib.nr`, `circuits/withdraw/`, `vks/`, `docs/DESIGN*.md`

brozorec:
Fixes #785 (audit finding N-07).

**The collision.** `encrypt_auditor_sender_balance` padded the withdraw balance checkpoint with `Poseidon2(AUDITOR_SENDER, [s_x, sigma])` — the *first* output of the sender-auditor sponge, which is exactly the transfer circuit's amount-pad slot. `transfer`, `set_spender`, and `revoke_spender` all already pad balances/allowances with the *second* squeeze; withdraw was the lone divergence. Under `(r_e, sigma)` reuse across a transfer and a later withdraw, anyone who knows the transfer amount (e.g. its recipient) can recover the shared pad from the published `v_tilde_aud_s` and decrypt the withdrawal's balance checkpoint.

**The fix.** The helper now returns `v_new + sponge_squeeze_2(AUDITOR_SENDER, s_x, sigma)[1]`, restoring a uniform slot assignment across every circuit: first squeeze = amount masks, second squeeze = balance/allowance masks. The checkpoint pad can no longer coincide with any amount pad, even under nonce reuse. This also makes the clawback decryption recipe in COMPLIANCE.md §5 (which already assumed the uniform two-squeeze form) correct for withdraw checkpoints.

Changes:
- `circuits/lib/src/lib.nr`: pad derivation + docstring (which also wrongly listed SetSpender/RevokeSpender as consumers — only withdraw calls this helper)
- `circuits/withdraw/`: W_a3 constraint comment; regenerated pinned fixture `B_TILDE_AUD_S`
- `vks/withdraw.vk.json`: regenerated with the pinned toolchain (only VK that changed); `constraints.baseline` is unchanged (same permutation count)
- `lib/testdata/encrypt_auditor_sender_balance.json`: regenerated vector + corrected description — **cross-language contract change**, downstream SDK/prover implementations of the withdraw checkpoint must switch to the second squeeze
- `docs/DESIGN.md`: W_a3 formula updated; §2.5 now states the canonical slot assignment and the normative requirement that a `(r_e, sigma)` pair be unique per proof, with the residual risk under reuse spelled out

Residual note: two balance checkpoints under a reused `(r_e, sigma)` still share their (second-slot) pad and leak their difference — inherent to any deterministic pad, hence the uniqueness requirement in §2.5.

Validation: `nargo test` passes across all circuit packages (153 tests); `cargo test -p stellar-tokens` (676) unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected auditor-sender withdrawal checkpoint encryption to use the designated second sponge output.
  * Updated withdrawal verification data and encrypted balance fixtures to match the corrected checkpoint behavior.

* **Documentation**
  * Clarified sponge slot usage for withdrawal checkpoints.
  * Added guidance requiring unique proof randomness to prevent mask reuse and potential ciphertext disclosure.

* **Tests**
  * Updated encryption, withdrawal, and fixture expectations for the revised checkpoint format.


ozgunozerk (approved):
Let put my understanding here, to confirm whether I got it correct. I'm approving in case I got it right. If my understanding is incorrect, please don't merge, I'll have to review again. 

My understanding:
- info is encrypted as: ciphertext = plaintext + pad -> one-time pad pattern
- pad is derived using nonce (technical detail, "how" part is not relevant to this PR)
- anyone who knows the plaintext and the ciphertext, can derive the pad. This means, the receiver of the transfer, can decipher to pad, due to they know they received `100` (plaintext), and ciphertext is already public info
- so, if the same pad is used elsewhere, considering the ciphertext is public info, they can decipher the plaintext. For example, the receiver can also decipher the remaining balance of the sender, which they shouldn't know ideally...

Pre-fix (before this PR), the same pad was used for 2 different things, potentially creating a private info leak, now different/unique pads are used for different inputs, fixing the problem

brozorec:
Yes, just to precise that it happens only if the same r_e and sigma are used for both, the transfer and the withdrawal.

Issue #33, opened 2025-02-05
🎁 [Feature Request]: Consider making `total_supply` tracking optional


dmkozh:
### What is the feature you would like to see?

SEP-41 doesn't require tokens to implement supply tracking, so it's not required for a the compliant tokens. Supply tracking has non-trivial performance cost, as its modification introduces a data dependency for all the transactions that use the token, which in turn makes them non-parallelizable, which reduces the token scalability.
It's hard to predict the exact impact - it depends on how often the token is being minted. But given the existence of the cost it would make sense to let the developer decide if they need this functionality given the potential performance cost.

brozorec:
Thanks for raising this question @dmkozh 🙏 

Indeed, SEP-41 doesn't require supply tracking for the reasons you mention, but they are specifically relevant for classic assets and their SACs. Custom tokens, which require more flexibility and because of that, won't be tied to a classic asset, can be SEP-41 compliant and "enjoy" additional functions.
Our goal by exposing a dedicated `FungibleToken` interface isn't to re-create/copy what's already available under `soroban_sdk::token::TokenInterface` but to give users more options for a "base" interface. Btw, `FungibleToken` is lacking `burn` and `burn_from` that must be implemented with `FungibleBurnable` for SEP-41 compliance.
We are providing [examples](https://github.com/OpenZeppelin/stellar-contracts/tree/main/examples/fungible-pausable/src) for both approaches when creating SEP-41 compliant tokens, explaining transparently the trade-offs.

dmkozh:
> Custom tokens, which require more flexibility and because of that, won't be tied to a classic asset, can be SEP-41 compliant and "enjoy" additional functions.

Sure, my issue here is rather that if a custom token doesn't need supply tracking, it's better off not supporting it, as it's not completely free. My point re SEP-41 was just to make it clear that it can be optional.

leighmcculloch:
The total supply field could be moved into an extension? `FungibleTotalSupply`

> its modification introduces a data dependency for all the transactions that use the token

Only transactions that mint and burn would need to write to the shared storage. So on the surface the impact depends on how, and how frequently, the token will be minted. Because tokens are a building block total supply has some risk of serialising txs that might otherwise have been independent.

PR #795, opened 2026-07-14
Feat/optional total supply

### LINKED ISSUE #560: 🏗️ [Core Feature]: Have a better architecture to enforce mutual exclusivity, without limiting composability and worsening DevX

ozgunozerk:
Fixes #560 

#### PR Checklist

- [x] Tests
- [x] Documentation

@brozorec, here is the PR that we were talking about, please take your time for reviewing it, because this is an important change. We should put our best effort into it to make it the "finalized" and "future proof" version if possible

## Summary by CodeRabbit

* **New Features**
  * Added opt-in total supply tracking for fungible tokens, including mint, burn, and supply queries.
  * Added composable token extensions for combining transfer policies with total supply tracking.
  * Added composition support for non-fungible token extensions.
  * Updated token examples to use the new composition and total supply capabilities.

* **Documentation**
  * Clarified extension behavior, supported combinations, and total supply requirements across token and vault examples.


bidzyyys:
## Review — stellar-tokens

Reviewed against the PR's own base, `origin/feat/contract-type-compose` (#821), at `9205bff` — line numbers will shift if #821 moves.

- [ ] 🔴 **packages/tokens/src/fungible/extensions/total_supply/storage.rs:166** — the supply moves from `instance`/`FungibleStorageKey::TotalSupply` (variant deleted at `fungible/storage.rs:33`) to `persistent`/`TotalSupplyStorageKey::TotalSupply`, so an already-deployed token that upgrades orphans its old entry and reads `0` — verified with an upgrade rehearsal: balances survive, `total_supply()` returns 0 — and for a vault that value feeds `convert_to_shares`/`convert_to_assets` (`vault/storage.rs:643,696`), silently mispricing shares instead of panicking → ship a migration helper (or read the legacy key as a fallback) and document the break, since removing `FungibleToken::total_supply` also drops an exported entry point.
- [x] 🟡 **packages/tokens/src/fungible/extensions/combinations/storage.rs:45-55** — `TotalSupplyBlockList`'s `transfer` / `transfer_from` / `approve` delegations are exercised by no test and no example (`fungible-blocklist` stays on `Compose<(BlockList,)>`), so a copy-paste slip routing them to `AllowList::*` would compile and pass CI while silently applying the inverted policy → mirror the three allowlist policy tests for `BlockListWithSupply`, or move the blocklist example to `Compose<(BlockList, TotalSupply)>`.
- [x] ⚪ **packages/tokens/README.md:21** — still advertises "total supply tracking" among the `fungible` module's capabilities, which the base trait no longer provides → point at the opt-in `total_supply` extension.
- [x] ⚪ **packages/tokens/src/fungible/extensions/total_supply/storage.rs:197** — `supply - amount` is unchecked with no `// NOTE: can't overflow because ...` proof, while this same PR replaced exactly that pattern in `Base::update` with `checked_add` → use `checked_sub` or add the proof comment anchored to the guards on lines 190/194.
- [x] ⚪ **packages/tokens/src/fungible/extensions/total_supply/storage.rs:159,189** — `increase_total_supply` / `decrease_total_supply` are public unauthenticated state writers whose `# Notes` omits the "does not enforce authorization" line their sibling `Base::update` carries → add it.
- [ ] ⚪ **commit 9205bff** — carries a `Co-Authored-By: Claude Fable 5` trailer, which `.claude/commands/code-quality.md` explicitly forbids → drop it on rebase.
- [ ] ⚪ **PR metadata** — the title is the raw branch name (`Feat/optional total supply`) rather than a summary like the commit's own "make fungible total supply tracking optional", and the body links `Fixes #560` while the commit says `Closes #33` (#560 is already claimed by #821) → fix the title and point the body at the issue this PR actually closes.

packages/tokens/src/fungible/extensions/votes/storage.rs
```diff
@@ -1,10 +1,23 @@
...
+// The voting checkpoints already track the total supply; the query is served
+// from them instead of a separate supply entry.
+impl TotalSupplyOverrides for FungibleVotes {
+    fn total_supply(e: &Env) -> i128 {
+        get_total_supply(e)
+            .try_into()
+            .unwrap_or_else(|_| panic_with_error!(e, FungibleTokenError::MathOverflow))
+    }
+}
+
```
brozorec: Although voting units correspond to token amounts in most of the cases, I believe we shouldn't assume they are equivalent. For a reference, the solidity [extension](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Votes.sol) doesn't override token's `totalSupply()`, it's actually another getter from `Votes` https://github.com/OpenZeppelin/openzeppelin-contracts/blob/6bc7bf38645e87468779c8118f6ba4b8bc6d4067/contracts/governance/utils/Votes.sol#L107 that accounts for voting units.

ozgunozerk: This will simplify it further, good news!

brozorec (approved):

PR #803, opened 2026-07-20
cannot remove/change identity with non-zero balance

### LINKED ISSUE #802: [M-03] - Identity repoint may invalidate compliances
Restrict identity repoint if the balance is > 0, also, get rid of `modify_identity`. One should use `remove` + `add` instead

ozgunozerk:
Fixes #802 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

- **Breaking Changes**
  - Identity records can no longer be modified through the identity registry storage API.
  - The related identity-modification event is no longer emitted.

- **Bug Fixes**
  - Identity removal now prevents deletion when the account holds a non-zero balance in any linked token.
  - Added a specific `AccountHasBalance` error for blocked removals.

- **Documentation**
  - Clarified identity-removal requirements, balance checks, and linked-token coverage.

### DIFF
--- a/packages/tokens/src/rwa/utils/token_binder/storage.rs
+++ b/packages/tokens/src/rwa/utils/token_binder/storage.rs
@@ -1,27 +1,33 @@
 use soroban_sdk::{contracttype, panic_with_error, Address, Env, Map, TryFromVal, Val, Vec};
 
 use crate::rwa::utils::token_binder::{
-    emit_token_bound, emit_token_unbound, TokenBinderError, BUCKET_SIZE, MAX_TOKENS,
-    TOKEN_BINDER_EXTEND_AMOUNT, TOKEN_BINDER_TTL_THRESHOLD,
+    emit_token_bound, emit_token_unbound, TokenBinderError, MAX_TOKENS, TOKEN_BINDER_EXTEND_AMOUNT,
+    TOKEN_BINDER_TTL_THRESHOLD,
 };
 
 /// Storage keys for the token binder system.
 ///
-/// - Tokens are stored in buckets of 100 addresses each
-/// - Each bucket is a `Vec<Address>` stored under its bucket index
-/// - Total count is tracked separately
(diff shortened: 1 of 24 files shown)


brozorec (commented):
Is it worth explaining that an identity modification, if needed, should be done in two steps, giving also some insights on why the interface diverges from the Solidity ref implementation?

Also given the bound tokens can be up to 10_000, we should make some estimates whether 10_000 calls fall within the budget. If not, we should reduce the limit.

bidzyyys:
## Review — identity_registry_storage (M-03)

- [x] 🟡 **packages/tokens/src/rwa/identity_verification/identity_registry_storage/storage.rs:446** — the balance-0 gate on `remove_identity` closes the `max_balance` vector but **not** `time_transfers_limits`: its per-identity `TransferCounter` tracks outbound velocity and is meaningful at balance 0, so `remove_identity` (bal 0) → `add_identity(account, new_identity)` still silently resets the transfer window — the exact velocity-reset M-03 names (and the "empty the account first" flow produces precisely this state) → also reconcile/gate identity-keyed velocity state on identity change (or migrate counters), not just balance; at minimum document that rotation must not circumvent open transfer windows.

packages/tokens/src/rwa/utils/token_binder/storage.rs
```diff
@@ -71,24 +67,11 @@ pub fn get_token_by_index(e: &Env, index: u32) -> Address {
 ///
 /// # Notes
 ///
-/// Performs a linear scan across all buckets. With Protocol 23, live state
-/// reads are inexpensive and read-entry limits have been removed.
+/// Performs a linear scan of the token list.
 pub fn get_token_index(e: &Env, token: &Address) -> u32 {
```
brozorec: after getting rid of the buckets, I'd say we don't need `get_token_index` and `get_token_by_index` anymore

brozorec (approved):

PR #810, opened 2026-07-21
eager prune for lockup

### LINKED ISSUE #809: [L-09] - lazy pruning for mint lockups, may grow too quickly
We can prune the expired lockups eagerly on every mint instead

ozgunozerk:
Fixes #809 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added a maximum capacity of 512 active lock entries per token and wallet.
  * Expired lock entries are automatically removed before new locks are recorded.
  * Added clear error reporting when minting or presetting locks exceeds the capacity limit.

* **Bug Fixes**
  * Prevented expired locks from incorrectly contributing to capacity limits.
  * Added coverage for lock pruning, capacity enforcement, and boundary conditions.

### DIFF
--- a/packages/tokens/src/rwa/compliance/modules/initial_lockup_period/storage.rs
+++ b/packages/tokens/src/rwa/compliance/modules/initial_lockup_period/storage.rs
@@ -541,10 +541,8 @@ fn prune_expired_locks(e: &Env, details: &mut LockedDetails) {
         }
     }
 
-    if pruned > 0 {
-        details.total_locked = sub_i128_or_panic(e, details.total_locked, pruned);
-        details.locks = remaining;
-    }
+    details.total_locked = sub_i128_or_panic(e, details.total_locked, pruned);
+    details.locks = remaining;
 }
 
 /// Consumes `amount` from the expired entries in `details`, oldest-first:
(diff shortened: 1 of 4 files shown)


bidzyyys:
## Review — stellar-tokens

- [x] ⚪ **src/rwa/compliance/modules/initial_lockup_period/storage.rs:269** — new doc claims the vector "cannot grow past the ledger-entry size limit under repeated issuance," but pruning only drops *expired* entries; many mints inside a single still-open lockup window still accumulate unbounded active locks (exactly the residual L-09's own text describes) → soften to "bounded by the number of concurrently-active locks" to match the accurate wording already used in mod.rs:42-45.

packages/tokens/src/rwa/compliance/modules/initial_lockup_period/storage.rs
```diff
@@ -289,6 +292,7 @@ pub fn on_created(e: &Env, to: &Address, amount: i128, token: &Address) {
     let period = get_lockup_period(e, token);
     if period > 0 && amount > 0 {
         let mut details = get_locked_details(e, token, to);
+        prune_expired_locks(e, &mut details);
         details.locks.push_back(LockedTokens {
```
brozorec, line 316: This follows the recommendation, but I'd suggest adding a hard cap. I guess there are cases when that exceeds the 128 KiB entry limit. Even though those can be considered edge cases, we should add a high-enough hard limit similarly to other places in the lib and ofc document it.

brozorec (changes requested):

brozorec (approved):

Issue #110, opened 2025-03-13
Add `event_assertion` to `non_fungible/test.rs`


ozgunozerk:
Currently, we have event assertions in `fungible`, but not in `non_fungible`.

PR #811, opened 2026-07-23
fix(sac-admin-generic): correct off-by-one in mint/clawback arg indices

brozorec:
## Summary

`MINT_AMOUNT_INDEX` and `CLAWBACK_AMOUNT_INDEX` in `sac_admin_generic/storage.rs` were `2`, counting the `e: Env` receiver as `args[0]`. But the `ContractContext.args` Soroban hands to `__check_auth` contains **only the invocation arguments** — the `Env` is never a wire argument. The real SAC layout is:

| fn | real `args` | correct index |
|---|---|---|
| `mint(to, amount)` | `[to, amount]` | amount = **1** |
| `clawback(from, amount)` | `[from, amount]` | amount = **1** |
| `set_authorized(addr, authorize)` | `[addr, authorize]` | bool = 1 (already correct) |

With the old value, `get_fn_param(ctx, 2)` did `args.get(2)` on a 2-element vec → `None` → `panic_with_error!(SACMissingFnParam)`. **Every `mint`/`clawback` routed through a generic SAC-admin contract panicked in `__check_auth`.** `set_authorized`/`set_admin` were unaffected.

The unit and example tests masked this by hand-building the context with a phantom leading `()` (`((), to, amount)`), so the assumed layout was never checked against a real SAC. They now model the real 2-argument layout, which also makes them a regression guard (setting the index back to `2` fails `test_extract_context_mint`).

## Verified live on Stellar testnet

Using a real deployed SAC + the generic-admin example contract, with a real ed25519-signed authorization entry (via the `sign-auth-ed25519` CLI plugin):

- **Before the fix** — tx `6466f553…` → FAILED; diagnostic trace: `SAC.mint → admin.__check_auth → error {contract: 111}` (SACMissingFnParam).
- **After the fix** — tx `fad1908d…` → SUCCESS; `__check_auth` returns, `mint` event emitted, balance credited `1000`.

## Checks

- `cargo +nightly fmt --all -- --check` ✅
- `cargo clippy --release --package stellar-tokens --package sac-admin-generic-example --all-targets -- -D warnings` ✅
- `cargo test` (tokens: 703 passed, example: 1 passed) ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected authorization handling for SAC mint and clawback operations.
  * Fixed amount extraction for contract invocation parameters.
  * Updated authorization validation for mint, clawback, and authorization changes.

* **Tests**
  * Added end-to-end coverage for minting, clawback, and authorization flows using real signed authorization entries.
  * Updated existing tests to reflect the corrected invocation argument format.

### DIFF
--- a/examples/sac-admin-generic/src/test.rs
+++ b/examples/sac-admin-generic/src/test.rs
@@ -111,7 +111,10 @@ fn test_sac_generic_e2e_mint_and_clawback() {
     // clawbackable.
     let initial_admin = Address::generate(&e);
     let sac = e.register_stellar_asset_contract_v2(initial_admin);
+    // `ClawbackEnabledFlag` makes minted balances clawbackable; `RevocableFlag`
+    // lets the admin deauthorize a balance via `set_authorized(_, false)`.
     sac.issuer().set_flag(IssuerFlags::ClawbackEnabledFlag);
+    sac.issuer().set_flag(IssuerFlags::RevocableFlag);
     let sac_client = StellarAssetClient::new(&e, &sac.address());
 
     // Deploy the generic SAC-admin contract (operator doubles as chief here).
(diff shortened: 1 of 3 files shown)


bidzyyys:
## Review — sac_admin_generic

- [x] ⚪ **examples/sac-admin-generic/src/test.rs:88** — e2e covers real-host parity for `mint`/`clawback` only; `SET_AUTHORIZED_BOOL_INDEX` is verified solely against a hand-built context → add a `build_auth("set_authorized", [addr, bool], nonce)` arm through the real SAC.
- [x] ⚪ **packages/tokens/src/fungible/utils/sac_admin_generic/test.rs:122** — mislabeled comment: `#[should_panic(#110)] // SACMissingFnParam`; `#110` is `SACAddressMismatch` → fix the comment.

ozgunozerk (approved):

PR #814, opened 2026-07-23
[N-13] - reduce duplicate calls 

### LINKED ISSUE #813: [N-13] - Redundant storage reads in burn/forced_transfer and duplicated identity verification in recover_balance
Two related inefficiencies in `packages/tokens/src/rwa/storage.rs`:

1. **Redundant storage reads:** `burn` and `privileged_transfer` (behind `forced_transfer`) capture the holder's balance and frozen amount in a pre-operation `AccountSnapshot`, then re-read the same two persistent slots via `get_free_tokens` / `get_frozen_tokens`, costing up to 3 redundant `get` + `extend_ttl` host calls per invocation on the unfreeze path even though nothing mutates the slots in between. Both values can be derived from the snapshot instead.

2. **Duplicated identity verification:** `recover_balance` verifies the new account's identity, then delegates to `privileged_transfer`, which runs `verify_identity` on the same address again, so every recovery pays for two full cross-contract verification cascades where one suffices. Moving the check out of `privileged_transfer` into `forced_transfer` makes the helper pure mechanics and leaves exactly one check per entry point.

ozgunozerk:
Fixes #813 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved identity verification during forced transfers.
  * Corrected token unfreezing calculations during privileged transfers and burns.
  * Burning tokens now properly reduces frozen balances when the burn exceeds the available free balance.

* **Tests**
  * Added coverage for burning tokens with automatic frozen-balance adjustments.

### DIFF
--- a/packages/tokens/src/rwa/storage.rs
+++ b/packages/tokens/src/rwa/storage.rs
@@ -224,6 +224,13 @@ impl RWA {
     /// **IMPORTANT**: This function bypasses authorization and freezing checks.
     /// Should only be used by authorized compliance or admin functions.
     pub fn forced_transfer(e: &Env, from: &Address, to: &Address, amount: i128) {
+        // Verify identity for the `to` address only; `from` is intentionally
+        // skipped so this privileged action can still move tokens out of
+        // accounts whose identity is no longer valid.
+        let identity_verifier_addr = Self::identity_verifier(e);
+        let identity_verifier_client = IdentityVerifierClient::new(e, &identity_verifier_addr);
+        identity_verifier_client.verify_identity(to);
+
         Self::privileged_transfer(e, from, to, amount, &TransferKind::Forced);
     }
 
@@ -234,6 +241,9 @@ impl RWA {
     /// `kind`. The two public entry points differ only in the kind they
     /// report, which is what lets compliance modules tell a seizure from a
     /// wallet migration.
+    ///
+    /// Performs no identity verification; each caller verifies the recipient
+    /// before delegating here, so the check runs exactly once per invocation.
     fn privileged_transfer(
         e: &Env,
         from: &Address,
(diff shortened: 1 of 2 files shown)


bidzyyys:
## Review — stellar-tokens (rwa)

Refactor is correct and behavior-preserving: `privileged_transfer`'s recipient check is now done by both callers (`forced_transfer` verifies `to`, `recover_balance` verifies `new_account`), snapshot-derived free/frozen values are algebraically identical to the removed re-reads, TTL extension is preserved, and no auth/reentrancy regression. fmt/clippy/wasm/tests(704)/coverage(97%) all green. Two minor items:

- [ ] ⚪ **src/rwa/storage.rs:475** — N-13's `recover_balance` case isn't closed: it still reads `Base::balance(old_account)` (:475) and `get_frozen_tokens(old_account)` (:481), then `privileged_transfer` re-`account_snapshot`s the same account (:257) with nothing mutating it in between → thread the snapshot (or its balance/frozen) into `privileged_transfer` so the source account is read once per recovery.
- [x] ⚪ **src/rwa/test.rs:321** — new `burn_with_token_unfreezing` asserts the frozen-balance delta but never pins the emitted `TokensUnfrozen` payload (no `.to_xdr` event comparison anywhere on the unfreeze path) → assert `TokensUnfrozen { user_address, amount: 30 }.to_xdr(&e, &address)` so a compliance indexer's wire contract is guarded.

brozorec (approved):
`recover_balance` still reads old_account's balance/frozen and then `privileged_transfer` re-snapshots the same account, but I'm fine to leave as-is

ozgunozerk:
> recover_balance still reads old_account's balance/frozen and then privileged_transfer re-snapshots the same account, but I'm fine to leave as-is

Yes, I left it deliberately, because it would require signature changes and potentially architectural changes. I think the minimal gains do not justify these changes.

PR #821, opened 2026-07-28
uniform contract type composition via Compose

### LINKED ISSUE #560: 🏗️ [Core Feature]: Have a better architecture to enforce mutual exclusivity, without limiting composability and worsening DevX

ozgunozerk:
The ContractType on FungibleToken and NonFungibleToken is now selected uniformly with the type-level Compose alias, by listing the contract types that override the Base behavior (e.g. Compose<(AllowList,)>). Invalid lists are rejected at compile time with a curated diagnostic. Additive extensions (burnable, capped, royalties) have no contract type and never appear in the list.

Also fixes the contract type bullet lists on both base traits: adds the missing Vault, FungibleVotes and NonFungibleVotes entries, and removes the incorrect claims that Enumerable/Consecutive are incompatible with NonFungibleBurnable.

Fixes #560 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added a public composition helper for selecting supported fungible- and non-fungible-token behaviors.
  * Improved compile-time diagnostics for unsupported or conflicting token extension combinations.
  * Updated token examples to use the new composition approach consistently.

* **Documentation**
  * Expanded guidance and examples for selecting and combining token behaviors.
  * Clarified extension precedence and mutually exclusive combinations.


packages/tokens/src/non_fungible/extensions/combinations/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+//! #[contractimpl(contracttrait)]
+//! impl NonFungibleEnumerable for MyToken {}
+//! ```
+//!
+//! No multi-type combinations are curated for non-fungible tokens yet; every
+//! valid list currently holds a single contract type. Invalid lists do not
+//! compile: `Enumerable` and `Consecutive` are mutually exclusive, and
+//! implementing an extension trait the list does not back (e.g.
+//! [`crate::non_fungible::enumerable::NonFungibleEnumerable`] without
+//! `Enumerable` in the list) is rejected by that trait's bound.
```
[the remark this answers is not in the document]
ozgunozerk: The PR removes the incorrect claims that Enumerable/Consecutive are incompatible with NonFungibleBurnable (both implement BurnableOverrides; both examples implement the trait). The Enumerable ↔ Consecutive mutual exclusivity is a different, retained claim and is correct by design: they compete for the single ContractType slot and their storage models conflict (Consecutive avoids per-token writes on batch mint, Enumerable requires them), so no curated combination can exist. This is also the long-standing framing in Architecture.md's mutual-exclusivity section. Unsupported-but-plausible combinations are documented as "not curated yet"; this pair is documented as exclusive because it's inherent, not pending.

bidzyyys:
## Review — stellar-tokens

- [x] 🟡 **packages/tokens/README.md:66,132** — the crate README still teaches the pre-`Compose` form `type ContractType = Base;` while `Architecture.md` and both base-trait docstrings now present `Compose<(...)>` as the way to select a contract type → migrate both snippets, plus the remaining mock sites in `packages/fee-abstraction/src/test.rs:40` and `examples/{fungible-vault,fungible-merkle-airdrop,fee-forwarder-permissioned,fee-forwarder-permissionless}/src/test.rs`.
- [x] ⚪ **packages/tokens/src/fungible/mod.rs:142** (and `non_fungible/mod.rs:130`) — the prose says to list "the contract types that override the `Base` behavior", but the vanilla selection is `Compose<(Base,)>`, which puts `Base` in the list of things overriding `Base` → reword to "the contract type providing the overridable behavior".
- [x] ⚪ **packages/tokens/src/non_fungible/mod.rs:105,110** — `Consecutive` and `Enumerable` are contract-type structs, but the new text calls each a "trait" (the old text named the actual traits `NonFungibleConsecutive` / `NonFungibleEnumerable`) → drop "trait" or point at the extension trait.
- [x] ⚪ **packages/tokens/src/fungible/extensions/burnable/mod.rs:38** — reflow left a ragged `and so on. There` line; the non-fungible sibling dropped the phrase instead → rewrap and make the two docstrings consistent.
- [x] ⚪ **packages/tokens/src/fungible/extensions/combinations/mod.rs** (and the non-fungible twin) — no `test.rs`, though the PR checklist marks Tests done; `(NonFungibleVotes,)` and all 10 bare non-tuple forms are exercised nowhere, so a mistyped `Out` in any of them would ship silently → add a small `test.rs` pinning each mapping (a compile-fail test for the invalid lists would need a harness the repo doesn't currently have).
- [x] ⚪ **commit dfd29e3** — carries a `Co-Authored-By: Claude Fable 5` trailer, which `.claude/commands/code-quality.md` explicitly forbids → drop it on rebase.

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+//! slot decides all overridable behavior. The contract type is selected
+//! uniformly with [`Compose`], by listing the contract types that override
+//! the `Base` behavior.
+//!
+//! Note that [`Compose`] only selects the `ContractType`; it says nothing
+//! about extensions that add new functionality without overriding the base
+//! behavior. Those (e.g. [`crate::non_fungible::burnable::NonFungibleBurnable`]
+//! or [`crate::non_fungible::royalties::NonFungibleRoyalties`]) have no
+//! contract type and are simply implemented alongside; there is no
+//! `Compose<(Burnable,)>`.
```
brozorec: I find this confusing 🤔 Even though `Burnable` and `Royalties` are not interfering with  the base behavior, shouldn't they be listed as possible combinations following the combinatorics exhaustivity claim?

(btw as currently implemented `Royalties` is not compatible with `Consecutive`: https://github.com/OpenZeppelin/stellar-contracts/pull/846)

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+impl Composable for Consecutive {
+    type Out = Consecutive;
+}
+impl Composable for (Consecutive,) {
+    type Out = Consecutive;
+}
+impl Composable for NonFungibleVotes {
+    type Out = NonFungibleVotes;
+}
+impl Composable for (NonFungibleVotes,) {
```
brozorec: I believe there shouldn't be an obstacle for `NonFungibleVotes` to be composed with either `Consecutive` or `Enumerable`. This is currently not possible, because it only overrides `Base`, but my understanding about the combinatorics pattern is that it is capable of solving this.

packages/tokens/src/fungible/extensions/combinations/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+/// Single contract types resolve to themselves (with or without the
+/// one-element tuple form). Mutually exclusive contract types (e.g.
+/// `AllowList` and `BlockList`) have no implementation, so combining them
+/// does not compile.
+#[diagnostic::on_unimplemented(
+    message = "`{Self}` is not a valid contract type combination",
+    note = "valid single contract types: `Base`, `AllowList`, `BlockList`, `RWA`, `Vault`, \
+            `FungibleVotes`",
+    note = "no multi-type combinations are curated for fungible tokens yet",
+    note = "`AllowList` and `BlockList` are mutually exclusive"
```
brozorec: I suggest re-evaluating this limitation as there could exist valid scenarios where they can be combined, e.g. one deploys a token with allowlisted users and throughout the token lifecycle another role can block/unblock some of the users.

packages/tokens/src/fungible/extensions/combinations/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+//! decides all overridable behavior. The contract type is selected uniformly
+//! with [`Compose`], by listing the contract types that override the `Base`
+//! behavior. [`Compose`] resolves the list to the curated contract type at
+//! compile time.
+//!
+//! Note that [`Compose`] only selects the `ContractType`; it says nothing
+//! about extensions that add new functionality without overriding the base
+//! behavior. Those (e.g. [`crate::fungible::burnable::FungibleBurnable`] or
+//! the [`crate::fungible::capped`] helpers) have no contract type and are
+//! simply implemented alongside; there is no `Compose<(Burnable,)>`.
```
brozorec: same note in https://github.com/OpenZeppelin/stellar-contracts/pull/821/changes#r3976762189

brozorec (changes requested):
I advocate for moving `allowlist` and `blocklist` to `contract-utils` crate so that they can be used in other contexts, e.g. `non-fungible`, `vault`, or non-related to the `tokens` crate.

packages/tokens/src/fungible/extensions/combinations/mod.rs
ozgunozerk: I'll address allowlist and blocklist abstraction and mutual exclusivity in another PR, it's not forgotten, don't worry 👍

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
ozgunozerk: Yes, good catch!

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
ozgunozerk: addressed in [09ff4c9](https://github.com/OpenZeppelin/stellar-contracts/pull/821/commits/09ff4c9ced9e505d30084ea7e1e8812b4381bab9)

packages/tokens/src/fungible/extensions/combinations/mod.rs
ozgunozerk: addressed in [09ff4c9](https://github.com/OpenZeppelin/stellar-contracts/pull/821/commits/09ff4c9ced9e505d30084ea7e1e8812b4381bab9)

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
ozgunozerk: will be addressed in another PR

packages/tokens/src/non_fungible/extensions/burnable/mod.rs
```diff
@@ -6,6 +6,15 @@ mod test;
...
 use soroban_sdk::{contractevent, contracttrait, Address, Env};
 
+/// Type-level name of the burnable extension, for use in a
+/// [`crate::non_fungible::combinations::Compose`] list.
+///
+/// Burnable is additive: it does not override the base behavior, so listing
+/// it is purely declarative and does not affect the resolved contract type.
+/// Burning is enabled by implementing [`NonFungibleBurnable`], whether or not
+/// `Burnable` is listed.
+pub enum Burnable {}
```
brozorec, line 16: is there a specific reason it's `enum` and not `struct`?

packages/tokens/src/non_fungible/extensions/combinations/mod.rs
brozorec, line 1: would it make sesne to make part of this plumbing common for fungible and non-fungible?

packages/tokens/src/non_fungible/extensions/burnable/mod.rs
ozgunozerk: Yes, `pub enum Burnable {}` -> so no value of it can ever exist. 

A unit struct (`pub struct Burnable;`) is itself a constructible value: 

`let b = Burnable;` compiles, it can be passed around and stored, which invites the question "what do I do with this?". 

The marker's entire role is to be a name in a Compose list; the empty enum turns "this is never used as a value" from a doc sentence into a compiler guarantee. The shape difference is also deliberately informative: contract types (Base, AllowList, Enumerable...) are unit structs because they carry associated functions, while additive markers carry nothing, and the enum/struct split mirrors that behavioral split. Switching to a struct would cost nothing functionally, it would just lose the can't-be-constructed guarantee.

brozorec (approved):

PR #841, opened 2026-08-17
Rwa batch

### LINKED ISSUE #767: RWA: Batch functions
Adds batch variants of the operator-gated RWA entry points so bulk operations across multiple accounts can be performed in a single call, mirroring the ERC-3643 standard's batch operations.

ozgunozerk:
Fixes #767 

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added atomic batch operations for RWA transfers, minting, burning, forced transfers, freezing, and unfreezing.
  * Added batch identity registration with input validation and event support.
  * Added a benchmark for measuring batch-operation resource usage and cost.

* **Documentation**
  * Added benchmark setup instructions, methodology, results, and resource-limit guidance.

* **Tests**
  * Expanded coverage for successful batch operations, validation errors, compliance checks, events, and verification behavior.


bidzyyys:
## Review — stellar-tokens

No High or Medium findings survived verification — the batch surface is fully guarded (`#[only_role]` on all six token batch methods and the registry's `batch_add_identity`), the sender-identity hoisting in `batch_transfer` is sound and pinned by an independent verifier-count oracle, and the "reorganization of storage" commit is a pure move (`privileged_transfer` is code-identical to `main`). Four low items:

- [x] ⚪ **bench/rwa-batch/Cargo.lock** — pins `soroban-sdk` 27.0.6 while the workspace lock pins 27.0.2, so the integrator-facing capacity figures published in `rwa/mod.rs` (~35 accounts for `batch_mint`/`batch_transfer`, ~35/99 for identity/freezing) were measured against a different SDK build than the library ships; the crate is workspace-`exclude`d so CI never rebuilds or re-runs it and the drift is silent → align the bench lock with the workspace SDK pin and re-measure before publishing the numbers.
- [x] ⚪ **packages/tokens/src/rwa/identity_verification/identity_registry_storage/mod.rs:419** — `batch_add_identity`'s doc puts a free paragraph ("The identity type is not part of the interface…") between `# Arguments` and `# Errors`, while its sibling `add_identity` keeps the block section-only → move it into the `# Notes` section already present below.
- [x] ⚪ **packages/tokens/src/rwa/test.rs** — `batch_transfer_credits_every_recipient` and `batch_burn_debits_every_account` assert only `events.events().last()`, so a batch emitting one event instead of one per item would still pass, whereas the sibling `batch_mint_credits_every_recipient` pins it properly with `events.events().len()` plus per-index `to_xdr` → follow the sibling's form in both.
- [ ] ⚪ **PR metadata / commits** — the title is the raw branch name ("Rwa batch") rather than a summary like the commits' own "batch variants of the operator-gated entry points", and three commits carry `Co-Authored-By: Claude Opus 5` trailers that `.claude/commands/code-quality.md` explicitly forbids → retitle and drop the trailers on rebase.

packages/tokens/src/rwa/storage.rs
```diff
@@ -31,6 +31,21 @@ pub enum RWAStorageKey {
 
 pub struct RWA;
 
+/// Who verifies the sender's identity in [`RWA::validate_transfer`].
+///
+/// Every item of a batch transfer shares one sender, so verifying it per item
+/// repeats the same walk through the identity stack. This lets the batch lift
+/// that one check out of its loop while every other check still runs per item,
+/// and while all three transfer entry points keep sharing one validator.
+pub enum SenderVerification {
```
brozorec: Given the owner of the tokens is designated as `from` and its related variables are `from_list` and `from_snapshot`, wouldn't be more coherent to name this enum `FromVerification`?

packages/tokens/src/rwa/storage.rs
```diff
@@ -835,4 +766,498 @@ impl RWA {
...
+    /// was just handed. A module granted write access to the identity stack
+    /// would break that assumption.
+    ///
+    /// The paused state and both wallets' freezing status are checked per item,
+    /// since those reads are cheap enough that hoisting them would buy nothing.
+    ///
+    /// The caller is responsible for sizing the batch to fit the
+    /// per-transaction network limits. Refer to the batch operations section of
+    /// the [module documentation](crate::rwa) for measured figures.
+    pub fn batch_transfer(e: &Env, from: &Address, recipients: &Vec<Address>, amounts: &Vec<i128>) {
```
brozorec, line 821: Compared to a regular transfer, here the order is `require_auth` → length check → `identity_verifier()` → full `verify_identity(from)` → `compliance()` → loop, and the paused check lives inside `validate_transfer`. That means that a paused token with an unverified sender now reports `IdentityVerificationFailed` instead of `EnforcedPause`. Also, an empty batch on a paused token succeeds where the single transfer panics.

brozorec (commented):
nice job 👍 ready to approve after we clear out those two comments

brozorec (approved):

PR #843, opened 2026-08-20
fix: preserve the SEP-41 muxed id in RWA transfer events

### LINKED ISSUE #842: RWA is the only `ContractType` that drops the SEP-41 muxed id from its transfer event
### What happens

`#646` added muxed transfer events to the fungible token, and `fungible/mod.rs` documents what they are for:

> Per SEP-41, when the destination is a `MuxedAddress` the event data carries both the amount and the muxed identifier **so that off-chain consumers can attribute the transfer to the correct sub-account.**

`Base::transfer` emits `to.id()` accordingly, and `AllowList`, `BlockList` and `FungibleVotes` all inherit that by delegating to `Base::transfer`.

RWA does not. Its `ContractOverrides::transfer` collapses the destination to its base address on the first line:

https://github.com/OpenZeppelin/stellar-contracts/blob/main/packages/tokens/src/rwa/storage.rs#L34-L36

```rust
impl ContractOverrides for RWA {
    fn transfer(e: &Env, from: &Address, to: &MuxedAddress, amount: i128) {
        RWA::transfer(e, from, &to.address(), amount);
    }
```

and every emit site inside RWA then passes `None` where the muxed id would go:

```rust
emit_transfer(e, from, to, None, amount);
```

`#646` touched `packages/test-utils/event-assertion/src/lib.rs` and `packages/tokens/src/fungible/{mod,storage,test}.rs` only, so RWA was never brought along.

### Reproduction

Two contracts wired identically with the canonical pattern from `examples/rwa/token/src/contract.rs`, differing only in `type ContractType`. Same `MuxedAddress` destination, muxed id 42, same amount.

```
CONTROL  ContractType = Base : V0(ContractEventV0 {
    topics: [Symbol(transfer), Address(Contract(...05)), Address(Account(...06))],
    data: Map(Some(ScMap([
        ScMapEntry { key: Symbol(amount),      val: I128(100) },
        ScMapEntry { key: Symbol(to_muxed_id), val: U64(42)   },
    ])))
})

SUBJECT  ContractType = RWA  : V0(ContractEventV0 {
    topics: [Symbol(transfer), Address(Contract(...05)), Address(Account(...06))],
    data: I128(100)
})

balances on the base account: base-token=100 rwa-token=100
```

The balances line is there to be explicit about scope. Nothing on chain diverges: the tokens land in the same place either way. The only difference is the event payload.

Verified against tag `v0.7.2` and against `main`, where the three `emit_transfer(e, from, to, None, amount)` sites are at `rwa/storage.rs` lines 288, 793 and 836.

### Why it matters

Anyone integrating an RWA token with a custodian, an exchange, or any deposit system that assigns sub-accounts by muxed id cannot attribute an incoming transfer. They get the base address and nothing else, which is exactly the problem `#646` and SEP-41 set out to solve for fungible tokens. Because the base token behaves correctly, the failure only appears once RWA is swapped in, which makes it easy to miss during integration.

There is no on chain consequence and no security impact that I can see. It is a conformance gap.

### The design question, and how the linked PR answers it

RWA's `ContractOverrides::transfer` narrows to `&Address` before any compliance or identity check runs, so threading the muxed id to the emit site means deciding whether the id should also be visible to `validate_transfer` and the compliance hook, or only to the event.

I have opened a PR alongside this issue so the change is concrete rather than described. It answers that question one particular way: the id reaches the event only, and identity, compliance and the balance update continue to see the base address. If you want it visible to the hook as well, that is a larger change and I would rather you decide it than have me assume.

Opening the PR alongside rather than after the discussion is only for concreteness. Happy to close it if you would rather settle the shape here first.

yigitcangokmen:
Fixes #842

### Problem

#646 added muxed transfer events to the fungible token so off-chain consumers can attribute a transfer to the correct sub-account. It touched `packages/tokens/src/fungible/` and the event-assertion helper only. `Base`, `AllowList`, `BlockList` and `FungibleVotes` all emit `to.id()`; RWA collapsed the destination to its base address on the first line of its `ContractOverrides::transfer` and passed `None` at every emit site.

### Change

The transfer body moves into a private `RWA::transfer_with_muxed_id` that takes `Option<u64>` and threads it to `emit_transfer`. `ContractOverrides::transfer` passes `to.id()`; the public `RWA::transfer` keeps its exact signature and passes `None`.

Identity verification, the compliance hook and `Base::update` all still receive the base `Address`.

For a muxed destination the emitted event does change wire shape, from `Transfer` with a bare `i128` payload to `MuxedTransfer` with a map payload. That is the intended fix, but it is downstream-visible and may warrant a release note. It only affects calls that pass a muxed address, which are mis-emitted today. That is a deliberate choice and it is the open design question here: if you want the muxed id visible to `validate_transfer` or to the compliance hook as well, that is a larger change and I would rather you decide it than assume.

`transfer_from`, `mint`, `burn`, `forced_transfer` and `recover_balance` all take a plain `Address` and are untouched; their emit sites still pass `None`, which is correct.

### Evidence

`contract_overrides_transfer_preserves_muxed_id` sends 30 units to a `MuxedAddress` with id 42 and asserts the emitted event.

Reverting only the source hunk and keeping the test gives:

```
left:  data: I128(30)
right: data: Map([ amount: I128(30), to_muxed_id: U64(42) ])
```

The test also asserts balances land on the base address, unchanged at 70 and 30, so the change is visibly confined to the event payload.

### Checks

- `cargo +stable clippy --release --locked --all-targets -- -D warnings` exit 0
- `cargo test -p stellar-tokens` passes: 714 tests, 713 before plus this one

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed RWA transfers to preserve destination IDs when sending to muxed addresses.
  * Transfer events now report the correct muxed destination ID.
  * Balances continue settling against the underlying base address.

### DIFF
--- a/packages/tokens/src/rwa/storage.rs
+++ b/packages/tokens/src/rwa/storage.rs
@@ -33,7 +33,7 @@ pub struct RWA;
 
 impl ContractOverrides for RWA {
     fn transfer(e: &Env, from: &Address, to: &MuxedAddress, amount: i128) {
-        RWA::transfer(e, from, &to.address(), amount);
+        RWA::transfer_with_muxed_id(e, from, &to.address(), to.id(), amount);
     }
 
     fn transfer_from(e: &Env, spender: &Address, from: &Address, to: &Address, amount: i128) {
@@ -771,6 +771,32 @@ impl RWA {
     /// Please refer to [`Base::update`] and [`Self::validate_transfer`] for the
     /// inline documentation.
     pub fn transfer(e: &Env, from: &Address, to: &Address, amount: i128) {
+        Self::transfer_with_muxed_id(e, from, to, None, amount);
+    }
+
+    /// `transfer` override that additionally carries the SEP-41 multiplexing
+    /// identifier of the destination into the emitted event.
+    ///
+    /// Identity verification, compliance and balance updates all operate on
+    /// the base `to` address; `to_muxed_id` only reaches the event, so that
+    /// off-chain consumers can attribute the transfer to the correct
+    /// sub-account.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `from` - The address holding the tokens.
+    /// * `to` - The base address receiving the tokens.
+    /// * `to_muxed_id` - The multiplexing identifier of the destination, if the
+    ///   caller supplied a muxed address.
+    /// * `amount` - The amount of tokens to be transferred.
+    fn transfer_with_muxed_id(
+        e: &Env,
+        from: &Address,
+        to: &Address,
+        to_muxed_id: Option<u64>,
+        amount: i128,
+    ) {
         from.require_auth();
 
         // Snapshot before the update so the hook observes the pre-transfer
@@ -790,7 +816,7 @@ impl RWA {
             &TransferKind::Standard,
             &e.current_contract_address(),
         );
-        emit_transfer(e, from, to, None, amount);
+        emit_transfer(e, from, to, to_muxed_id, amount);
     }
 
     /// `transfer_from` override with added compliance and identity verification
--- a/packages/tokens/src/rwa/test.rs
+++ b/packages/tokens/src/rwa/test.rs
@@ -2,13 +2,13 @@ extern crate std;
 
 use soroban_sdk::{
     contract, contractimpl, panic_with_error, symbol_short,
-    testutils::{Address as _, Events},
-    Address, Env, Event, String,
+    testutils::{Address as _, Events, MuxedAddress as _},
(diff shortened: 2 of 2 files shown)


brozorec:
@ozgunozerk can you plz check here. What are the implication of this PR with regards to SEP-57? Given a RWA is a fungible token, should we adopt the same stance about `MuxedAddress` as in fungible?

ozgunozerk:
Thanks a lot for this PR!

`RWA::transfer()` needs the signature change instead of a helper function. 

Closing this one in favor of #857, and will give you contributor credits there as `Co-Author` due to your contributions 🥳

PR #844, opened 2026-08-20
fix(fee-abstraction): drop Lazy-mode expiration check that validates the wrong value

### LINKED ISSUE #840: fee-abstraction: Lazy-mode expiration check added in #546 validates the wrong value, not the actual allowance
## Description

In `Lazy` approval mode, `collect_fee()` (`packages/fee-abstraction/src/storage.rs:188-201`) only re-approves when the existing allowance is insufficient:

```rust
FeeAbstractionApproval::Lazy => {
    let allowance = token_client.allowance(user, &e.current_contract_address());
    if allowance < max_fee_amount {
        token_client.approve(user, &e.current_contract_address(), &max_fee_amount, &expiration_ledger);
    } else {
        // assuming that in the other cases the expiration ledger is validated in `token.approve()`
        validate_expiration_ledger(e, expiration_ledger);
    }
}
```

When the existing allowance is already sufficient, `approve()` is never called, so `expiration_ledger` is never applied to anything on-chain: the real, operative allowance keeps whatever expiration a *previous* `approve()` call set. `validate_expiration_ledger(e, expiration_ledger)` only checks that the *parameter passed to this call* is in the future, which has no relationship to the real allowance's actual expiration.

This line was added deliberately in #546, described as *"Added explicit expiration ledger validation for the case we don't rely on `token.approve()`"*. Checking the wrong variable means it doesn't accomplish that in either direction:

1. **False rejection.** A fee collection against a genuinely still-valid allowance can panic if the caller passes a smaller `expiration_ledger` than the current ledger sequence, even though nothing is actually expired. Demonstrated by the crate's own test, `collect_fee_with_lazy_approval_expired_ledger_panics` (`packages/fee-abstraction/src/test.rs:198-229`): the real allowance is approved for 100 units until ledger 200, current ledger is set to 101, `max_fee_amount` is 50 (well within the real allowance), and the call still panics with `#5006 InvalidExpirationLedger` purely because the test passes `expiration_ledger: 100 < 101`.
2. **No real tightening.** Conversely, a caller trying to *shorten* an existing long-lived allowance by passing an earlier `expiration_ledger` doesn't shrink anything: the prior, longer expiration silently remains what's actually enforced.

The check doesn't add protection the SAC's own `transfer_from` expiry enforcement wasn't already providing, while adding a real false-rejection surface.

## Suggested direction

Either drop the check in the `else` branch (the SAC's own expiry enforcement at `transfer_from` already covers the real allowance), or, if the intent is for `Lazy` mode to let a caller tighten an existing approval's expiration without bumping the amount, that needs an actual `approve()` call when the expiration should change, not a validation of an unused parameter.

## Verified against

`main` at `fbfde38` (current HEAD). The existing test suite passes as written and demonstrates the behavior above; `collect_fee_with_lazy_approval_expired_ledger_panics` isn't wrong about what the code does, it's confirming the gap.

Eras256:
## Root cause, confirmed independently

`collect_fee()`'s `Lazy` branch only re-approves when the existing allowance is already below `max_fee_amount`. When it isn't, the `else` branch ran `validate_expiration_ledger(e, expiration_ledger)` — which only checks that the *parameter passed to this call* is in the future. That parameter has no relationship to the real, on-chain allowance's actual expiration: it was already fixed by whichever prior `approve()` call established it, and `TokenClient` doesn't even expose it as a readable value (only the amount, via `allowance()`).

Ran the crate's own cited test (`collect_fee_with_lazy_approval_expired_ledger_panics`) against `main` at `fbfde38` before touching anything: it panics with `Error(Contract, #5006)`, and the diagnostic event log shows `allowance: 100` at the moment of the panic — confirming the real allowance was genuinely sufficient and non-expired when the call was rejected. Then wrote two new, independent tests to isolate both directions the issue describes before writing this fix:

- A real allowance of 100 units valid until ledger 200, current ledger 101 (99 ledgers of real validity left), still panics on `expiration_ledger: 100 < 101` — a false rejection.
- An `expiration_ledger` of 9999, nowhere near the real allowance's actual expiration of 200, is accepted just as readily as any other value — the parameter never reaches the real allowance either way.

Both confirmed the bug is real and exactly as scoped, not an edge case or a misreading of the original report.

## Fix (option a from the issue)

The issue's own "Suggested direction" left two paths open:

- **(a)** drop the check in the `else` branch — the SAC's own `transfer_from` expiry enforcement, called two lines later in the same function, already protects the real allowance; this check wasn't adding coverage `transfer_from` didn't already provide, only a false-rejection surface.
- **(b)** if `Lazy` mode is meant to let a caller *tighten* an existing approval's expiration without bumping the amount, that needs an actual `approve()` call when the expiration should change, not a validation of an unused parameter.

This PR implements **(a)**, the smaller and less debatable of the two — dropping the check entirely rather than changing `Lazy` mode's semantics. `validate_expiration_ledger` itself is untouched and stays public (it still has its own direct unit test, `validate_expiration_ledger_past`); this PR only removes the one call site that was checking it against the wrong value. If the maintainers want **(b)** instead — real tightening via a conditional `approve()` — happy to rework this PR for that instead; flagging it here since it's a real behavior fork, not just an implementation detail.

## Tests

- `cargo test --lib` in `packages/fee-abstraction`: 22/22 passing (20 pre-existing + 2 new).
- `cargo clippy --all-targets --all-features -- -D warnings`: clean.
- `cargo +nightly fmt --all -- --check`: clean.
- The old `collect_fee_with_lazy_approval_expired_ledger_panics` test encoded the bug as expected behavior (`#[should_panic]`). Replaced with `collect_fee_with_lazy_approval_succeeds_regardless_of_expiration_ledger_param` (same exact setup, now asserts success and the correct post-call allowance) and `collect_fee_with_lazy_approval_ignores_unrelated_expiration_ledger_param` (the second direction — an arbitrary future param has no real effect).

Fixes #840.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fee collection now succeeds when an existing allowance covers the fee, regardless of the supplied expiration ledger parameter.
  * Prevented valid allowances from being incorrectly rejected due to unrelated or outdated expiration values.

* **Tests**
  * Added coverage confirming successful fee collection and correct allowance and recipient balances in these scenarios.

### DIFF
--- a/packages/fee-abstraction/src/storage.rs
+++ b/packages/fee-abstraction/src/storage.rs
@@ -194,11 +194,13 @@ pub fn collect_fee(
                     &max_fee_amount,
                     &expiration_ledger,
                 );
-            } else {
-                // assuming that in the other cases the expiration ledger is validated in
-                // `token.approve()`
-                validate_expiration_ledger(e, expiration_ledger);
             }
+            // Else: the existing allowance already covers max_fee_amount, so
+            // no approve() call is issued here, and `expiration_ledger` is
+            // never applied to anything. The real allowance's expiration was
+            // already set by whichever prior approve() call established it,
+            // and the SAC enforces that on `transfer_from` below — there is
+            // nothing left to validate `expiration_ledger` against.
         }
     }
 
--- a/packages/fee-abstraction/src/test.rs
+++ b/packages/fee-abstraction/src/test.rs
@@ -195,9 +195,20 @@ fn collect_fee_with_lazy_approval_lower_previous() {
     assert_eq!(balance, 20);
 }
 
+// Regression test for #840. The `expiration_ledger` parameter has no
+// relationship to the real, on-chain allowance's actual expiration — that
+// expiration was already fixed by whichever prior approve() call set it, and
+// TokenClient doesn't even expose it as a readable value, only the amount via
+// `allowance()`. Before the fix, this exact scenario (a genuinely valid,
+// non-expired allowance of 100 until ledger 200, current ledger 101) panicked
+// with `Error(Contract, #5006)` purely because the *unrelated* parameter
+// (100) happened to be less than the current ledger sequence (101) — a false
+// rejection against a real allowance that still had 99 ledgers of validity
+// left. After the fix, the call succeeds: the SAC's own `transfer_from`
+// expiry enforcement is what actually protects the real allowance, and it's
+// not fooled by this parameter either way.
 #[test]
-#[should_panic(expected = "Error(Contract, #5006)")]
-fn collect_fee_with_lazy_approval_expired_ledger_panics() {
+fn collect_fee_with_lazy_approval_succeeds_regardless_of_expiration_ledger_param() {
     let e = Env::default();
     e.mock_all_auths_allowing_non_root_auth();
 
@@ -209,7 +220,7 @@ fn collect_fee_with_lazy_approval_expired_ledger_panics() {
     let max_fee_amount = 50;
 
     let token_client = TokenClient::new(&e, &token_address);
-    // approve enough (100 > max_fee_amount) till ledger 200
+    // approve enough (100 > max_fee_amount) till ledger 200 — genuinely valid
     token_client.approve(&user, &contract_address, &100, &200);
 
     e.ledger().set_sequence_number(101);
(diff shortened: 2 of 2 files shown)


Eras256:
Following up here — this has been mergeable and CI-green since Aug 24 (`b68a9ab`), just waiting on a review. @brozorec, could you take a look when you have a moment? Happy to address anything that comes up.

brozorec:
closed with https://github.com/OpenZeppelin/stellar-contracts/issues/840

PR #846, opened 2026-08-20
fix: dispatch royalty_info existence check through ContractType

### LINKED ISSUE #845: `royalty_info` hardcodes `Base::`, so it reverts for 9 of every 10 tokens under `ContractType = Consecutive`
### What happens

`NonFungibleRoyalties::royalty_info` calls `Base::royalty_info` directly instead of dispatching through `Self::ContractType`:

https://github.com/OpenZeppelin/stellar-contracts/blob/main/packages/tokens/src/non_fungible/extensions/royalties/mod.rs#L144-L146

```rust
fn royalty_info(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
    Base::royalty_info(e, token_id, sale_price)
}
```

`Base::royalty_info` establishes existence with `Base::owner_of`:

```rust
pub fn royalty_info(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
    // Verify token exists by checking owner
    let _ = Base::owner_of(e, token_id);
```

Under `Consecutive`, ownership is stored sparsely: only the batch boundary token gets a materialised `Owner` entry, and `Consecutive::owner_of` walks back through the bucket to resolve the rest. `Base::owner_of` does a direct lookup, so it panics `#200 NonExistentToken` for every token that is not a boundary.

The result is that a contract using `Consecutive` together with the royalties extension has a `royalty_info` that reverts for almost every token it owns.

### This deviates from the library's own documented convention

`packages/tokens/src/non_fungible/overrides.rs#L21` states that default bodies should call `Self::ContractType::{function_name}`, and every sibling follows it:

```rust
// non_fungible/mod.rs
fn balance(e: &Env, account: Address) -> u32     { Self::ContractType::balance(e, &account) }
fn owner_of(e: &Env, token_id: u32) -> Address   { Self::ContractType::owner_of(e, token_id) }
fn transfer(e: &Env, from: Address, ...)         { Self::ContractType::transfer(e, &from, &to, token_id); }
```

`royalty_info` is the exception. It looks like it was missed when the default bodies were added in #604.

### Reproduction

Two contracts wired identically, differing only in `ContractType`. Full source at the bottom.

```
CONTROL  ContractType = Base        : 10/10 token ids answer royalty_info
SUBJECT  ContractType = Consecutive : 10/10 token ids answer owner_of
SUBJECT  ContractType = Consecutive :  1/10 token ids answer royalty_info
         failing token ids: [0, 1, 2, 3, 4, 5, 6, 7, 8]
```

The middle line is the important control. `Consecutive::owner_of` resolves all ten tokens correctly, so the tokens exist and the extension is working as designed. Only `royalty_info` fails, and it fails because of how it dispatches, not because of anything wrong with the token.

Only token 9 succeeds, because 9 is the batch boundary and therefore the one id with a materialised `Owner` entry.

Verified against tag `v0.7.2` and against `main` (the line is unchanged on `main`).

### The fix, and the design question in it

The straightforward fix routes the existence check through `Self::ContractType::owner_of`, which is how every other default body in this library dispatches, and moves the lookup into a `Base::royalty_info_unchecked` so `Base::royalty_info` behaves exactly as before.

That adds public API to a published crate, so it is a real decision rather than a mechanical change. The alternative is to declare `royalty_info` on `ContractOverrides` with the `Base` behaviour as the default and override it in `Consecutive`, which grows the trait instead of the `Base` surface. I have a working patch for the first shape and will open it as a PR against this issue, but I do not have a strong view and would rather you chose.

yigitcangokmen:
Fixes #845

### Problem

`NonFungibleRoyalties::royalty_info`'s default body called `Base::royalty_info`, which establishes existence with `Base::owner_of`. Under `ContractType = Consecutive` ownership is stored sparsely: only the last token of a batch has a materialised `Owner` entry, and the rest are resolved by walking back through the bucket. Every non-boundary token was therefore rejected with `#200 NonExistentToken`.

The library documents the correct pattern in `non_fungible/overrides.rs`, and every sibling default body follows it. `royalty_info` was the exception, which looks like it was missed when the default bodies landed in #604.

### Change

The existence check now goes through `Self::ContractType::owner_of`, matching how every other default body in this library dispatches.

The lookup itself moves into `Base::royalty_info_unchecked`, so `Base::royalty_info` keeps its current behaviour exactly: it still calls `Base::owner_of` and then the same lookup.

**This adds public API to a published crate**, which is worth flagging up front. The `_unchecked` suffix follows `credit_identity_unchecked` in `rwa/compliance/modules/max_balance/storage.rs`, and the new function carries a `# Security Warning` section as that precedent does. If you would rather not grow the surface, the alternative is to put `royalty_info` on `ContractOverrides` with the `Base` behaviour as the default and override it in `Consecutive`. Say the word and I will redo it that way.

### Evidence

`test_royalty_info_resolves_every_token_under_consecutive` batch-mints ten tokens and asserts that each one resolves both an owner and royalty information.

Reverting only the two source hunks and keeping the test gives:

```
test ...resolves_every_token_under_consecutive ... FAILED
HostError: Error(Contract, #200)
  data: ["contract call failed", royalty_info, [0, 1000000]]
```

The test also asserts `owner_of` answers for all ten. That is the control: the tokens demonstrably exist and `Consecutive` resolves them, so the failure is in the dispatch rather than the token.

### Checks

- `cargo +stable clippy --release --locked --all-targets -- -D warnings` exit 0
- `cargo test -p stellar-tokens` passes: 714 tests, 713 before plus this one. `nft-royalties-example` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Royalty information now works correctly for tokens issued through consecutive token batches.
  * Royalty lookups validate token ownership before returning royalty details.

* **Tests**
  * Added regression coverage for royalty resolution across all tokens in a consecutive batch.

### DIFF
--- a/packages/tokens/src/non_fungible/extensions/royalties/mod.rs
+++ b/packages/tokens/src/non_fungible/extensions/royalties/mod.rs
@@ -1,5 +1,5 @@
 mod storage;
-use crate::non_fungible::{Base, NonFungibleToken};
+use crate::non_fungible::{Base, ContractOverrides, NonFungibleToken};
 
 #[cfg(test)]
 mod test;
@@ -142,7 +142,13 @@ pub trait NonFungibleRoyalties: NonFungibleToken {
     /// * [`crate::non_fungible::NonFungibleTokenError::NonExistentToken`] - If
     ///   the token does not exist.
     fn royalty_info(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
-        Base::royalty_info(e, token_id, sale_price)
+        // Establish existence through the contract's own ownership model.
+        // `Base::owner_of` only resolves tokens that have a materialised
+        // `Owner` entry, so calling it directly would reject every token that
+        // `Consecutive` resolves by walking back through its bucket.
+        let _ = Self::ContractType::owner_of(e, token_id);
+
+        Base::royalty_info_unchecked(e, token_id, sale_price)
     }
 }
 
--- a/packages/tokens/src/non_fungible/extensions/royalties/storage.rs
+++ b/packages/tokens/src/non_fungible/extensions/royalties/storage.rs
@@ -159,6 +159,29 @@ impl Base {
         // Verify token exists by checking owner
         let _ = Base::owner_of(e, token_id);
 
+        Self::royalty_info_unchecked(e, token_id, sale_price)
+    }
+
+    /// Returns the royalty information for a token **without** verifying that
+    /// the token exists.
+    ///
+    /// # Arguments
+    ///
+    /// * `e` - Access to the Soroban environment.
+    /// * `token_id` - The identifier of the token.
+    /// * `sale_price` - The sale price for which royalties are being
+    ///   calculated.
+    ///
+    /// # Security Warning
+    ///
+    /// **IMPORTANT**: This function does not verify that the token exists.
+    /// Callers must establish existence first, using the ownership model that
+    /// applies to their contract. [`Base::royalty_info`] does so with
+    /// [`Base::owner_of`]; a contract whose `ContractType` stores ownership
+    /// differently, such as
+    /// [`crate::non_fungible::consecutive::Consecutive`], must use its own
+    /// `owner_of` instead.
+    pub fn royalty_info_unchecked(e: &Env, token_id: u32, sale_price: i128) -> (Address, i128) {
         // Check if there's a specific royalty for this token
         let token_key = NFTRoyaltiesStorageKey::TokenRoyalty(token_id);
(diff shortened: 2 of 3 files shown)


ozgunozerk:
@yigitcangokmen thanks for the issue, and the PR. 

The solution needs to be more robust, and route itself automatically based on the underlying trait. Also, not only for the `royalty_info`, but all the other operations. 

Since we cannot put `ContractType::owner_of` in the main trait here, due to we deliberately lack default implementations, we had to find another way. The required solution must be completely different than this approach proposed in this PR, hence closing in favor of #879

PR #848, opened 2026-08-20
fix: cap spending history at the reachable ledger entry limit

### LINKED ISSUE #847: `MAX_HISTORY_ENTRIES` is unreachable, so `HistoryCapacityExceeded` can never be returned
### Summary

`spending_limit` caps its history vector to guard against storage growth:

```rust
/// Maximum number of spending entries to keep in history.
/// This prevents storage DoS by capping the vector size.
pub const MAX_HISTORY_ENTRIES: u32 = 1000;
```

```rust
if data.spending_history.len() >= MAX_HISTORY_ENTRIES {
    panic_with_error!(e, SpendingLimitError::HistoryCapacityExceeded)
}
```

The ledger entry stops being writable before that check can fire. A `SpendingLimitData` holding **816** entries serializes to **65,592 bytes**, past the mainnet `contractDataEntrySizeBytes` limit of 65,536. The measured ceiling is **815**. So instead of the typed error the policy documents, a caller at the ceiling gets an untyped host budget failure 185 entries early.

### Measurement

Driving the real `enforce` with resource limits at their mainnet defaults, which `Env::default()` applies automatically:

```
invocation resource limits are exceeded: contract data entry with key
'ContractData(... Symbol(AccountContext), ..., durability: Persistent)'
size: 65592 > 65536

Error(Budget, ExceededLimit)
```

Filling to 815 succeeds; 816 fails. Both boundaries are pinned by two tests, which I will open as a PR against this issue.

### Why no existing test catches it

`enforce_history_capacity_exceeded` reaches 1000 because it calls `e.cost_estimate().disable_resource_limits()`. To be clear, that is not a mistake: the call is a normal idiom in this repo, used in eight places across four files for tests that deliberately exceed mainnet limits. The gap is narrower than that. No test currently covers the interaction between this constant and the platform limit, so the two can drift apart silently.

### Where 1000 came from

The cap was the accepted fix for **M-09** in the v0.5.0 audit. That report sized the entry as:

> "This single ledger entry is subject to a 128 KiB size limit. Each `SpendingEntry` item has a raw size of 20 bytes (`i128` for amount is 16 bytes, `u32` for `ledger_sequence` is 4 bytes). Thus, the maximum number of spending entries is approximately: 131,072 bytes (128 KiB) / 20 bytes/entry ≈ 6,553 entries."

1000 was a comfortable margin under 6,553. Two things moved since:

1. The applicable limit is `max_contract_data_entry_size_bytes`, **65,536**, not 131,072. 131,072 is `max_contract_code_entry_size_bytes`.
2. A `SpendingEntry` is a `#[contracttype]` struct, so it encodes as an `SCV_MAP` with symbol keys at roughly **80 bytes** on the wire rather than the raw field width of 20.

Together that is 2x from the limit and 4x from the entry size, which is the whole distance between 6,553 and 815.

### The question

**What should the cap be?**

I have deliberately left the constant alone, for two reasons. Lowering it is a behavioural change for any account currently holding more than the new value inside its rolling window, and this is an audited security module where the headroom judgement is yours rather than mine.

The PR documents the ceiling and adds tests that pin both sides of it, so whatever you choose stops being able to drift silently. If you would rather compare against a serialized size than a count, or store the history more compactly than a `Vec` of `#[contracttype]` structs, I am happy to send that instead.

### Scope

This is not a security report. `enforce` opens with `smart_account.require_auth()` and rejects an empty signer set, so the party who fills the history is an authorised signer of the very context rule they render unusable. Recovery is automatic once entries age out of the rolling window, and the effect is scoped to one rule by the `AccountContext(smart_account, rule.id)` key. In both worlds the account is blocked once it saturates; what changes is whether the failure is diagnosable.

yigitcangokmen:
Fixes #847

### What this does

Caps `spending_history` at the value that is actually reachable, so the guard fires and callers get `HistoryCapacityExceeded` instead of an untyped host budget error.

```diff
-pub const MAX_HISTORY_ENTRIES: u32 = 1000;
+pub const MAX_HISTORY_ENTRIES: u32 = 815;
```

### Why 815

A `SpendingLimitData` holding 816 entries serializes to 65,592 bytes, past the mainnet `contractDataEntrySizeBytes` of 65,536. At 1000 the guard could never fire: `enforce` died with `Error(Budget, ExceededLimit)` 185 entries early and `HistoryCapacityExceeded` was unreachable.

### On `disable_resource_limits`

`enforce_history_capacity_exceeded` was only green because it called `e.cost_estimate().disable_resource_limits()`. With limits off the entry size ceiling never applies, which is probably why the unreachable cap went unnoticed. I dropped that one line, so the test now fills the history under mainnet defaults and still gets `#3224`. The other 7 uses of the idiom elsewhere in the repo are untouched. Can put it back if you'd rather keep it.

### Tests

- `enforce_history_capacity_exceeded` now runs under mainnet limits and still expects `#3224`.
- `spending_history_past_the_cap_exceeds_the_ledger_entry_limit` writes one entry past the cap straight to storage, bypassing the guard, so the ceiling is pinned by a test rather than a comment. It asserts `Error(Budget, ExceededLimit)` rather than a contract error code, which is deliberate: the failure genuinely is a host rejection and that is the point being pinned.

### Checks

- `cargo clippy -p stellar-accounts --all-targets -- -D warnings` exit 0
- `cargo test -p stellar-accounts`: 182 passed, 0 failed

### DIFF
--- a/packages/accounts/src/policies/spending_limit.rs
+++ b/packages/accounts/src/policies/spending_limit.rs
@@ -158,14 +158,15 @@ pub const SPENDING_LIMIT_TTL_THRESHOLD: u32 = SPENDING_LIMIT_EXTEND_AMOUNT - DAY
 ///
 /// # Notes
 ///
-/// This cap is currently **unreachable**. A `SpendingLimitData` holding 816
-/// entries serializes to 65,592 bytes, past the mainnet
-/// `contractDataEntrySizeBytes` limit of 65,536, so [`enforce`] fails with a
-/// host budget error at 816 and [`SpendingLimitError::HistoryCapacityExceeded`]
-/// is never returned. The measured ceiling is 815 entries.
-/// `spending_history_ledger_entry_ceiling` in the test module pins both
-/// boundaries.
-pub const MAX_HISTORY_ENTRIES: u32 = 1000;
+/// The cap is the largest history that still fits in a single ledger entry. A
+/// `SpendingLimitData` holding 816 entries serializes to 65,592 bytes, past the
+/// mainnet `contractDataEntrySizeBytes` limit of 65,536, so a higher cap could
+/// never be enforced: [`enforce`] would fail with an untyped host budget error
+/// before [`SpendingLimitError::HistoryCapacityExceeded`] could be returned.
+/// Raising this value therefore requires a smaller entry or a larger network
+/// limit, which
+/// `spending_history_past_the_cap_exceeds_the_ledger_entry_limit` pins.
+pub const MAX_HISTORY_ENTRIES: u32 = 815;
 
 // ################## QUERY STATE ##################
 
--- a/packages/accounts/src/policies/spending_limit.rs
+++ b/packages/accounts/src/policies/spending_limit.rs
@@ -155,6 +155,16 @@ pub const SPENDING_LIMIT_TTL_THRESHOLD: u32 = SPENDING_LIMIT_EXTEND_AMOUNT - DAY
 
 /// Maximum number of spending entries to keep in history.
 /// This prevents storage DoS by capping the vector size.
+///
+/// # Notes
+///
+/// This cap is currently **unreachable**. A `SpendingLimitData` holding 816
+/// entries serializes to 65,592 bytes, past the mainnet
+/// `contractDataEntrySizeBytes` limit of 65,536, so [`enforce`] fails with a
+/// host budget error at 816 and [`SpendingLimitError::HistoryCapacityExceeded`]
+/// is never returned. The measured ceiling is 815 entries.
+/// `spending_history_ledger_entry_ceiling` in the test module pins both
+/// boundaries.
 pub const MAX_HISTORY_ENTRIES: u32 = 1000;
 
 // ################## QUERY STATE ##################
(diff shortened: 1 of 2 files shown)


brozorec:
Hey @yigitcangokmen good catch and thanks for this PR. I suggest we set the limit to what's reachable, i.e. 815.

yigitcangokmen:
Thanks @brozorec  🦆🦆

Done, `MAX_HISTORY_ENTRIES` is 815 now:

```diff
-pub const MAX_HISTORY_ENTRIES: u32 = 1000;
+pub const MAX_HISTORY_ENTRIES: u32 = 815;
```

The guard is reachable at that value, so you get `HistoryCapacityExceeded` instead of a budget error.

**Heads up on one gotcha though.** `enforce_history_capacity_exceeded` was only green because of this:

```rust
e.cost_estimate().disable_resource_limits();
```

With limits off the entry size ceiling never kicks in, which is probably why nobody caught that the cap was unreachable. I dropped the line and the test still gets `#3224`, just under mainnet limits now. Can put it back if you'd rather keep it.

Also kept one test that writes 816 entries straight to storage, bypassing the guard, so the ceiling is pinned by a test instead of a comment:

<details>
<summary>host output at 816 entries</summary>

```
contract data entry with key 'ContractData(...)' size: 65592 > 65536
HostError: Error(Budget, ExceededLimit)
```

</details>

`cargo test -p stellar-accounts`: 182 passed, 0 failed.

brozorec (approved):
LGTM

PR #853, opened 2026-08-26
feat(confidential): escrow commitment blindings to the sender-auditor

### LINKED ISSUE #764: Confidential Token: Seize flow (Clawback)
Implements the admin + auditor coordinated seizure flow specified in COMPLIANCE.md §5: a dedicated circuit and contract entry point that let the admin freeze a target account and settle a proven amount to the issuer, using a zero-knowledge proof (bounded by the auditor's off-chain-recovered spendable + receiving balance) so the seized amount stays hidden from anyone but the two coordinating parties.

brozorec:
**Stack 1/3.** Base: `main`. Followed by [proofless revoke](../../tree/feat/conf-token-proofless-revoke) → [clawback](../../tree/feat/conf-token-clawback).

## What

Widens the sender-auditor Poseidon2 sponge from two lanes to three, and adds an auditor-side escrow of the allowance blinding.

- `sponge_squeeze_3` in `circuits/lib/src/lib.nr`
- Lane 2 carries **the blinding of a commitment the operation writes, never a key** — the new spendable blinding on the checkpoint operations (**W_a5**, **T_a9**, **S_a6**), the new allowance blinding `r_a'` on spender transfers (**O_a9**)
- **S14** escrows `r_a` to the owner's auditor at `set_spender`, under new domain tag 17 (`ESCROWED_ALLOWANCE_BLINDING_AUDITOR`) — a single-output pad, because `set_spender`'s lane 2 is already taken by S_a6

Net effect: the auditor recovers the full Pedersen *opening* of `C_spend` and `C_a`, not just the value.

## Scope of the capability

The escrow is of a per-state blinding, not of `dvk_i`. That matters because `dvk_i = Poseidon(δ_dvk, vk, op_i)` is permanent per `(owner, spender)` — no salt or generation counter enters it, so revoke-then-re-delegate regenerates it exactly — and `σ_a` is public. Escrowing it would make one leaked ciphertext open every allowance state for that pair, past and future, with neither rotation nor revocation remediating.

So decryption is **event-scoped**: a key can decrypt exactly the ciphertexts produced while it was active. A rotated-in key holds no opening for a live delegation until that delegation's next state change. It does *not* follow that a retired key loses track of post-rotation state — proofless folds recompute `C_spend' = C_spend + C_a` publicly with no new encrypted checkpoint, so a holder of a known opening carries it forward by the same addition the contract performs. `DESIGN_cont.md` §8.5 states this and claims nothing stronger.

**Tradeoff:** the auditor must have observed the event; there is no cold state-based recovery. This extends an existing dependency rather than creating one (DESIGN §5.2 already makes a durable event archive normative, and every other auditor capability is event-scoped). New `INDEXER.md` §7.1 covers the consequences — a fails-closed check of each reconstructed opening against the stored `allowance_commitment`, and delegation-entry TTL as the bound on how long that check stays possible.

## Why it's safe to widen

`sponge_squeeze_3(d, s, σ)[0..2] == sponge_squeeze_2(d, s, σ)` by construction — the absorb fits one rate-3 block, so both read the same permutation. **No existing mask value changes**, and the widening costs no extra constraints. Pinned by `sponge_squeeze_3_agrees_with_squeeze_2_on_shared_lanes`.

Tag 17 is separate from `ESCROWED_DELEGATION_VIEWING_KEY` (10) even though reuse would be secure: sharing it would drop one leg of DESIGN §5.3's "why reusing `r_e` is safe" argument. S14 is the thinnest construction in the system — it reuses the S_a2 shared scalar and absorbs `op_i` (a per-delegation constant, not a nonce), so it stands on tag distinctness plus the freshness of `s_{a,s}` alone. §5.3 now says so plainly.

## Review notes

Read `sponge_squeeze_3` once, then check four near-identical applications. Constraint counts against `main`:

| Circuit | ACIR before | after |
|:--|--:|--:|
| Withdraw | 94 | 95 |
| Transfer | 133 | 134 |
| SetSpender | 131 | 135 |
| SpenderTransfer | 135 | 136 |

Both escrowed blindings were already constrained witnesses in scope above their escrow sites (S6/S7, O10/O11), so S14 costs one Poseidon and O_a9 one field addition.

One thing that reads as a mismatch and is not: O_a9's lane-2 pad keys off the *consumed* `σ_a` (the whole sponge does, per O_a6) while the escrowed value derives from the *written* `σ_a'`. Sound — the pad needs only to be fresh and unpredictable to everyone but the auditor — and `DESIGN_cont` §8.4 states it explicitly.

Payloads gain `r_tilde_aud_s` (all four operations) and `r_a_tilde_aud_s` (set_spender); PI order and event shapes follow. Four VKs regenerate. **This is a wire-format change** — mixed old/new proofs and verification keys fail cleanly; treat as an SDK migration and call it out in release notes.

New tests worth a look: `rejects_escrow_of_blinding_for_the_wrong_sigma_a` and `rejects_escrow_of_the_consumed_allowance_blinding` pin that each circuit escrows the blinding of the state it *writes*; five characterization tests document what happens under `σ_a` / `r_e` reuse (including the A→B→A alternation O10 permits) rather than asserting it is prevented, since nothing enforces global uniqueness; and `auditor_key_rotation_rescopes_the_escrowed_allowance_opening` walks set → rotate → spender transfer → revoke, asserting which key version each operation's ciphertexts were produced for.

## Checks

`nargo test` 169 passed · `cargo test -p stellar-tokens` 714 passed · `cargo +nightly fmt --check` clean · `cargo clippy --release --locked --all-targets -D warnings` clean · `constraints.baseline` regenerated · VKs regenerated


packages/tokens/src/confidential/circuits/lib/src/lib.nr
```diff
@@ -316,27 +373,43 @@ pub fn encrypt_esc_dvk(dvk: Field, s: Field, op_i: Field) -> Field {
...
-pub fn encrypt_auditor_sender_balance(v_new: Field, s_a_s: Field, sigma: Field) -> Field {
-    v_new + sponge_squeeze_2(domain::AUDITOR_SENDER, s_a_s, sigma)[1]
+/// `sponge_squeeze_3(d, s, sigma)[0..2] == sponge_squeeze_2(d, s, sigma)`, by
+/// construction: the absorb fits one rate-3 block, so all three outputs are
+/// served from the same permutation and the two functions differ only in how
+/// much of the state they expose. Widening therefore changes no existing mask
+/// value and costs no extra constraints. The capacity lane at `state[3]` stays
+/// untouched.
+///
+/// The blindings lane 2 carries still belong to different commitment roles --
```
ozgunozerk: i'd avoid using language such as `lane 1` or `lane 2` because it is prone to confusion between human counting or computer counting. 

`lane 2` -> second lane? or `lane[2]` ?

I think in this one it is referring to the 3rd lane, which is `lane[2]` but better to be explicit throughout this all PR about it

packages/tokens/src/confidential/storage.rs
```diff
@@ -141,6 +144,9 @@ pub struct TransferPayload {
     pub r_tilde_aud_r: BytesN<32>,
     pub v_tilde_aud_s: BytesN<32>,
     pub b_tilde_aud_s: BytesN<32>,
+    /// Sender-auditor secret-escrow slot (sponge lane 2): the blinding of the
+    /// sender's new spendable commitment, `r_A' + m_r_s` (constraint T_a9).
+    pub r_tilde_aud_s: BytesN<32>,
```
ozgunozerk: this is commented because it is newly written, but what about the above variable under this struct? Similarly for the other places/structs this new field is added. Having only one field commented and other ones not does not look consistent

ozgunozerk (approved):

ozgunozerk (commented):
Only 2 comments that I couldn't find the answers on my own:

1. σ_a on SetSpender and σ_a' on SpenderTransfer are new event fields, and the description never mentions them. Neither is needed by the decryption paths §8.5 specifies (S14's pad keys off op_i, not σ_a). Presumably they're there to back §8.5's "reconstructed from the event alone with no storage read". Ask why, and whether the auditor actually needs them or whether it's for the spender/indexer.

2. Pad reuse now costs more than it used to. r_e = Poseidon(δ_eph, vk_op, σ_a) is derived from the consumed salt, and spender_transfer/src/main.nr:276 only asserts sigma_a_new != sigma_a, so an A→B→A alternation reproduces a pad exactly. Previously a repeated pad leaked a value; now lane 2 makes it leak the other half of an opening. Nothing enforces global salt uniqueness and the PR characterizes this in tests (alternating_sigma_a_reproduces_an_earlier_allowance_blinding, repeated_r_e_and_sigma_a_repeat_every_channel_pad) rather than preventing it. That's a defensible call, but it's the sharpest security question here and I'd want the docs to say the blast radius grew.

PR #857, opened 2026-08-27
Rwa muxed support

### LINKED ISSUE #842: RWA is the only `ContractType` that drops the SEP-41 muxed id from its transfer event

ozgunozerk:
Fixes #842 

#### PR Checklist

- [x] Tests
- [x] Documentation

Credit is shared with @yigitcangokmen, will add him as a Co-Author when this PR is to be merged.

## Summary by CodeRabbit

* **New Features**
  * Added support for transfers to muxed addresses while preserving their destination identifiers in transfer events.
  * Balances and compliance checks continue to apply to the underlying address.
* **Bug Fixes**
  * Prevented issuers that do not support custodial sub-accounts from accepting muxed destinations.
  * Maintained enforcement of freeze and identity-verification rules for muxed transfers.
* **Tests**
  * Added coverage for muxed transfer events, aggregate balances, compliance restrictions, and unsupported destination handling.


brozorec (commented):
Should we also propose an update on SEP-57?

ozgunozerk:
Definitely! I was waiting for a review first before suggesting any change on SEP57

brozorec (approved):

PR #867, opened 2026-09-03
feat(examples): add smart account factory example

### LINKED ISSUE #861: Add a smart account factory
### Problem

A Soroban address derives from (network, deployer, salt). A smart account's signers and policies are constructor arguments, so they are not in the address.

With a G… deployer whose secret is publicly derivable (as in smart-account-kit), anyone can compute the predicted C… and create there first with a different signer set. The victim's create fails, funds sent to the predicted address land on the attacker's contract, and an attacker who deploys the real wasm with both parties as signers still passes an "is my key a signer?" check.

### Proposal

A factory contract at "examples/multisig-smart-account/factory", referenced by address like the verifier and policy examples, following the "soroban-examples/deployer" pattern:

- Only the factory can create in its namespace.
- The chain salt hashes the account constructor arguments actually deployed (signers list, policies map) with a caller-chosen intermediate salt, so a mismatched constructor cannot land at the predicted address. The caller-chosen part exists only to allow several accounts with identical constructor arguments.
- Generic over the full constructor, not a pinned signer type or wasm hash.

ericglau:
Fixes #861

Adds a new example contract, `examples/multisig-smart-account/factory`, providing a permissionless deploy factory for the multisig smart account in `examples/multisig-smart-account/account`. Like the verifier and policy contracts, one factory instance is deployed per network and referenced by address.

### Interface

```rust
__constructor(account_wasm_hash: BytesN<32>)

predict_address(signers: Vec<Signer>, policies: Map<Address, Val>, salt: u32) -> Address
deploy(signers: Vec<Signer>, policies: Map<Address, Val>, salt: u32) -> Address

pinned_account_wasm_hash() -> BytesN<32>
```

### Design Decisions

- **Namespace Exclusivity:** Accounts are deployed using `e.deployer().with_current_contract(chain_salt).deploy_v2(...)`. On Stellar, contract addresses derive from `(network, deployer, salt)`, so only this factory can deploy into its namespace.
- **Constructor Binding:** The `chain_salt` is computed as `sha256(XDR((signers, policies, salt)))`. Duplicate signer identities are rejected with `DuplicateSigner` (same check as the account, including External keys that canonicalize equal). The remaining list is sorted through a host `Map`, and the policy map is a host-ordered `Map`. The exact canonical values hashed into the salt are passed directly to `deploy_v2`, so only the predicted configuration can be created at the predicted address.
- **Intermediate Salt Discriminator:** The caller supplies an intermediate `salt: u32` rather than a raw 32-byte salt. This allows the same configuration of signers and policies to deploy multiple distinct accounts, while ensuring the factory alone derives the final 32-byte `chain_salt` that binds constructor arguments.
- **Account Wasm Pinning:** The account wasm hash is pinned on the factory instance during `__constructor`. Because code hash is not part of a Stellar contract address, pinning the wasm at factory construction prevents arbitrary bytecode from landing at a predicted address. Deploying a different account wasm would require deploying a separate factory instance.
- **Generic Constructor Passthrough:** Takes `signers: Vec<Signer>` and `policies: Map<Address, Val>` without hardcoded signer limits or module allowlists.

### Documentation & Walkthrough

A CLI walkthrough is included in `examples/multisig-smart-account/factory/README.md`, following the setup and key conventions of the parent multisig example.

### Verification

Tests run against the account wasm compiled from `examples/multisig-smart-account/account` (stored under `testdata/`, matching the pattern in `examples/upgradeable/testdata`):

```bash
cargo +nightly fmt --all -- --check
cargo clippy --release --locked --all-targets --package multisig-account-factory-example -- -D warnings
cargo test --package multisig-account-factory-example
cargo doc --no-deps --package multisig-account-factory-example
stellar contract build --package multisig-account-factory-example
```

#### PR Checklist

- [x] Tests
- [x] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added a permissionless factory for deploying multisig smart accounts.
  * Added deterministic address prediction based on signers, policies, and a salt.
  * Added support for deploying multiple accounts with distinct salts and preventing duplicate deployments.
  * Added documentation covering factory setup, WASM pinning, address prediction, and deployment workflows.

* **Tests**
  * Added coverage for signer and policy normalization, deployment behavior, address derivation, and error handling.


examples/multisig-smart-account/factory/README.md
```diff
@@ -0,0 +1,113 @@
+# Smart Account Factory Example
+
+Use this factory if you want to deploy accounts to deterministic addresses
+derived from their initial signers and policies, or predict an address before
+deployment.
+
+This guide deploys the multisig smart account from the
+[account example](../account) through a factory. Like the verifier and policy
```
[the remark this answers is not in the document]
ericglau: `../account` is the account crate this factory deploys, not the walkthrough. The guide is already linked in the following list and in Setup.

examples/multisig-smart-account/factory/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## 5. Deploy the Account
+
+Deploy with the same tuple. The returned address is the one `predict_address` gave:
+
+```bash
+stellar contract invoke --id account_factory -- deploy \
+    --signers '[ ...same as above... ]' \
+    --policies '{ ...same as above... }' \
```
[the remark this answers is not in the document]
ericglau: The example JSON is shown once when setting `SIGNERS` and `POLICIES`. Both `predict_address` and `deploy` reuse those variables so the commands stay valid and cannot drift.

examples/multisig-smart-account/factory/src/test.rs
```diff
@@ -0,0 +1,287 @@
...
+fn signer_order_and_duplicates_do_not_change_the_address() {
+    let e = Env::default();
+    let (factory, verifier, _) = setup(&e);
+    let client = AccountFactoryContractClient::new(&e, &factory);
+    let a = external(&e, &verifier, 1);
+    let b = external(&e, &verifier, 2);
+
+    let ab = client.predict_address(&vec![&e, a.clone(), b.clone()], &no_policies(&e), &0);
+    let ba = client.predict_address(&vec![&e, b.clone(), a.clone()], &no_policies(&e), &0);
+    assert_eq!(ab, ba);
```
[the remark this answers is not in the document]
ericglau: Fixed in `08b4213` by renaming those locals to `order_first_second` / `order_second_first` (and the duplicate-signer pair to `duplicated` / `single`).

examples/multisig-smart-account/factory/src/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+        let wasm_hash = Self::pinned_account_wasm_hash(e);
+        let signers = canonical_signers(e, &signers);
+        let chain_salt = chain_salt(e, &signers, &policies, salt);
+
+        e.deployer().with_current_contract(chain_salt).deploy_v2(wasm_hash, (signers, policies))
+    }
+}
+
+/// Sorts and deduplicates signers through a host `Map`.
+fn canonical_signers(e: &Env, signers: &Vec<Signer>) -> Vec<Signer> {
```
brozorec, line 91: Would it make sense following signers' canonicalization check from [here](https://github.com/OpenZeppelin/stellar-contracts/blob/6ea3075d44de51ea2a0d26fa45cbee6d385731b5/packages/accounts/src/smart_account/storage.rs#L543)? Especially for `External` signers.

ericglau: Thanks — `canonical_signers` now calls `validate_no_canonical_duplicates` on the input list first (exact and External canonical duplicates fail with `DuplicateSigner`), then sorts via a host `Map` so signer order still does not change the address. We no longer drop duplicates.

brozorec (approved):

examples/multisig-smart-account/factory/README.md
```diff
@@ -0,0 +1,118 @@
...
+than the transaction caller, the account address is independent of who submits
+or pays for the deployment transaction. The 32-byte `chain_salt` is computed as
+`sha256` of the canonical XDR of `(signers, policies, salt)`, where `signers`
+and `policies` are the account constructor arguments and `salt` is a
+caller-chosen `u32`.
+
+The account wasm hash is pinned on the factory instance during
+`__constructor`. Because code hash is not part of a Stellar contract address,
+pinning the wasm at factory construction prevents arbitrary bytecode from
+landing at a predicted address.
```
ozgunozerk: Can we better explain what is the problem here that is getting resolved? And also how? 

After reading these intro, it is still unclear to me

ozgunozerk (approved):

examples/multisig-smart-account/factory/README.md
ozgunozerk: Thanks!

ericglau: Thanks. Rewrote part of the intro with **The problem** / **How this factory solves it**. Derivation is based on `(network, deployer, salt)`; without a factory that salt is whatever the deployer picks, so constructor arguments and wasm never enter it. This factory is the deployer, derives the salt from the constructor arguments, and pins the wasm.

PR #873, opened 2026-09-09
fee-abstraction: consume the whole eager approval

brozorec:
fix https://github.com/OpenZeppelin/stellar-contracts/issues/875

## Summary by CodeRabbit

- **Behavior Changes**
  - Eager fee collection now uses the full approved maximum, pays the actual fee, refunds any remainder, and clears the allowance.
  - Eager collection requires sufficient token balance for the approved maximum and fails otherwise.
  - Lazy fee collection continues to transfer only the actual fee and preserves any unused allowance.

- **Documentation**
  - Clarified approval, refund, allowance, balance, and failure behavior for fee collection.

Credits go to @knQzx for identifying the issue and proposing a fix

### DIFF
--- a/examples/fee-forwarder-permissionless/src/contract.rs
+++ b/examples/fee-forwarder-permissionless/src/contract.rs
@@ -45,8 +45,10 @@
 //!
 //! 6. **Contract executes atomically**:
 //!    - Validates both user and relayer authorizations
-//!    - User approves contract to spend up to `max_fee_amount` tokens
-//!    - Contract transfers exactly `fee_amount` tokens from user to itself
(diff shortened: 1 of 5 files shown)


ozgunozerk (changes requested):
I think 2 other PRs are included in this PR, I don't think it is intentional. 

I'm talking about:
- `frozen` status carry over fix for RWA
- and spending limit constant change and test addition from external contributor

brozorec:
> I think 2 other PRs are included in this PR, I don't think it is intentional.

Yes, it was initially targeting `main` and that's why.

ozgunozerk (approved):

PR #895, opened 2026-09-15
vault virtual supply bound

### LINKED ISSUE #894: 🐞 [Bug]: Vault: a mint can leave total_supply + 10^offset unrepresentable, freezing all conversions
### What happened?

Every vault conversion computes total_supply + 10^offset (the real shares plus the virtual shares) with a checked add and panics with MathOverflow if the sum does not fit in i128. The mint path, however, only checks that total_supply + shares fits. So a deposit or mint can succeed and leave the stored supply less than 10^offset below i128::MAX. From then on every deposit, mint, withdraw, redeem and preview fails on that addition. Since both exit paths convert before burning or transferring, deposited assets cannot leave through the vault interface.

The state is only reachable at the numeric ceiling: it requires a single vault, at the maximum offset of 10, to hold about i128::MAX / 10^10 base units of the underlying, roughly 17 billion whole tokens of an 18-decimal asset. The poisoning deposit is the last base unit of remaining capacity, so this is a robustness gap rather than a practical attack, and we classify it as Low. Fix: check total_supply + shares + 10^offset in deposit_internal before the asset transfer, which covers both deposit and mint. Reported through the bug bounty program.

### Expected behavior

_No response_

ozgunozerk:
Fixes #894 

#### PR Checklist

- [x] Tests
- [x] Documentation

We should talk about which milestone this issue should be assigned to as well


packages/tokens/src/vault/storage.rs
```diff
@@ -79,6 +79,13 @@ pub enum VaultStorageKey {
 /// assets will cause the first user to exit to experience reduced losses in
 /// detriment to the last users that will experience bigger losses.
 ///
+/// The virtual shares also lower the ceiling of the share supply cap. Every
+/// conversion computes `total_supply + 10^offset`, so the real supply has to
+/// stay at least `10^offset` below `i128::MAX` for the vault to keep working.
+/// [`Vault::deposit_internal()`] enforces this bound before minting. Contracts
+/// that mint shares through any other path (for example, a mintable extension
+/// on the share token) are expected to preserve it themselves.
```
brozorec: I suggest we drop the example with the mintable extension as we don't provide such, just keep the general statement.

packages/tokens/src/vault/storage.rs
```diff
@@ -730,6 +739,12 @@ impl Vault {
     /// * `from` - The address that will provide the underlying assets.
     /// * `operator` - The address performing the deposit operation.
     ///
+    /// # Errors
```
brozorec, line 742: the errors from `query_asset`, `increase_total_supply` and  `Base::update` are missing

packages/tokens/src/vault/storage.rs
```diff
@@ -730,6 +739,12 @@ impl Vault {
...
+    ///
+    /// * [`VaultTokenError::MathOverflow`] - When minting `shares` would push
+    ///   the share supply plus the virtual shares (`10^offset`) past
+    ///   `i128::MAX`, which would make every later conversion fail.
```
brozorec:
```suggestion
    /// * [`VaultTokenError::MathOverflow`] - When minting `shares` would push
    ///   the share supply plus the virtual shares (`10^offset`) past
    ///   `i128::MAX`.
```

brozorec (approved):

